PR #1, opened 2023-02-09
Add ADR for naming themes

gbirke:
This is for https://phabricator.wikimedia.org/T327623

### DIFF
--- a/docs/adr/002-theme-names.md
+++ b/docs/adr/002-theme-names.md
@@ -1,0 +1,105 @@
+# Theme names
+
+2023-02-05
+
+## Status
+
+Accepted
+
+## Context
+
+A "theme" for a banner groups all the styling (colors, fonts, margins, etc) related to each
+component in files named after the components. The theme (and its corresponding folder) needs to
+have a name.
+
+We came up with a list of must-have requirements for theme names:
+
+- everyone should be able to come up with a new theme name
+- the name should be short
+- the name should be to write and pronounce
+- the list of names should be consistent (i.e. follow the same pattern or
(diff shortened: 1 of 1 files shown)


moiikana:
@ everyone:
I'd like to propose to have 1 name for mobile and desktop themes that share the same appearance.

e.g. "funnyWhiteOrangeTheme" (desktop) and "funnyWhiteOrangeTheme_mobile" for the mobile banner that uses the same color scheme / main elements
Talking about the design ( at least team internal ) might be easier with less new names to handle.
Please let me know what you think

Sperling-0:
> @ everyone: I'd like to propose to have 1 name for mobile and desktop themes that share the same appearance.
> 
> e.g. `funnyWhiteOrangeTheme` (desktop) and `funnyWhiteOrangeTheme_mobile` for the mobile banner that uses the same color scheme / main elements.
> 
> Talking about the design ( at least team internal ) might be easier with less new names to handle. Please let me know what you think

I agree with @moiikana here. Also, as we have only two options for our theme names, I feel using the 'creative animal' names for our themes would be better. As we won't have to remember new words (which are dummy).

Abban:
We just need to make sure the names we pick have as few syllables as possible, so maybe we need to audit that list before generating?

gbirke:
Hi @wmde/funtech-core I've adapted the ADR to your suggestions (adding suffixes, using the random pronounceable word generator because it can generate short 6-character words) and added the word list. Feel free to use the comment/suggest function to strike words you can't pronounce or don't like.

Abban (approved):

PR #18, opened 2023-03-07
Banner Transitions

Abban:


.eslintrc.cjs
```diff
@@ -43,6 +43,10 @@ module.exports = {
 			style: { lang: 'scss' }
 		} ],
 
-		'@typescript-eslint/type-annotation-spacing': 'error'
+		'@typescript-eslint/type-annotation-spacing': 'error',
+
+		// ESLint treats Typescript enums weird
```
gbirke, line 48:
```suggestion
		// Workaround for Typescript enums, see https://github.com/typescript-eslint/typescript-eslint/issues/2483
```
When I saw this I had no context, was baffled and needed to search for "eslint typescript enums"

src/components/BannerConductor/BannerConductor.vue
```diff
@@ -0,0 +1,62 @@
...
+const bannerState = ref<BannerState>( BannerState.Pending );
+
+onMounted( () => {
+	props.page.setTransitionDuration( props.bannerConfig.transitionDuration )
+		.setSpace( bannerRef.value.offsetHeight );
+
+	setTimeout( () => {
+		bannerState.value = BannerState.Showing;
+		props.page.setAnimated()
+			.showBanner();
```
gbirke, line 39: From the `BannerConductor` name I guess you'll add more code here?

At the end of the timeout it should
1. Check if banner should/can be displayed
2. Do some event tracking (if it was not displayed, log reason (size issues,  etc) and notify CN see https://phabricator.wikimedia.org/T310982 )
3. set animated/shown properties if it should be displayed.

.stylelintrc.json
```diff
@@ -31,6 +31,14 @@
...
 				]
 			}
+		],
+		"no-invalid-position-at-import-rule": [
+			true,
+			{
+				"ignoreAtRules": [
+					"use"
+				]
+			}
```
gbirke, line 41: We might not need this rule if we use `@use` consistently

src/components/BannerConductor/StateMachine/states/ClosedState.ts
```diff
@@ -0,0 +1,19 @@
+import { BannerState } from '@src/components/BannerConductor/StateMachine/states/BannerState';
+import { BannerStates } from '@src/components/BannerConductor/StateMachine/BannerStates';
+
+export class ClosedState extends BannerState {
```
gbirke, line 4: The way the "Closed" state and state machine is set up currently, it'll probably encapsulate the tracking for "closing" the banner. Keep in mind that with the current setup we'll have different "events" that lead to a closed state:
- Size issue
- User interaction with the page (i.e. page object instance) before banner is displayed
- User interaction (with different close buttons / soft close, etc)

We could also split this state to make it more meaningful, into `Closed` (through direct user interaction with banner elements) and `Canceled` (page object prevented banner display). The `Canceled` state could then call the `mw.centralNotice.setBannerLoadedButHidden('reason')` method required by https://phabricator.wikimedia.org/T310982 
The `Canceled` constructor would also take the page object like the `Pending` one does, but we wouldn't need it for the `Closed` one, which is a good thing because the page object should be constrained to the `BannerConductor` (which is also the the one that makes the transition to `Canceled` 


The more I think of it, the more I like this state machine, because its constructor will force us to think about events when closing (not showing) the banner. 
We'll have to discuss how we handle the different tracking events (`BannerEvent` vs `SizeIssue`) and how (or if) we pass the event handler into the constructors. 

I'd like the API on the *calling* side to be as clear and lean as possible, which is why I'd go with some kind of factory method

```typescript
// examples of different events
bannerState.changeState(CloseState.newForBannerSizeIssue());
bannerState.changeState(CloseState.newForUserInteraction('soft-close-timeout'))
```

We'd need to figure out how we can share the event handling instance among the states without needing to pass it into every constructor. Maybe pass it into `BannerStateMachine` and pass it as a parameter to the `enter` and `exit` methods?

src/components/BannerConductor/StateMachine/states/ClosedState.ts
```diff
@@ -1,11 +1,19 @@
...
+	page: Page;
+
+	constructor( page: Page ) {
+		super();
+		this.page = page;
+	}
 
 	enter(): Promise<any> {
-		// fire events
+		this.page.unsetAnimated();
```
Abban, line 16: We should add a method to the page called `removeBanner()` or `hideBanner()` that also stops resize event listeners

Abban: Oh, and calls `mw.centralNotice.hideBanner` thorought the MediaWiki object

PR #19, opened 2023-03-08
Add loading spinner for compile and copy actions on Banner dashboard

Sperling-0:
Ticket: https://phabricator.wikimedia.org/T328551

### DIFF
--- a/dashboard/components/BannerActions.vue
+++ b/dashboard/components/BannerActions.vue
@@ -20,7 +20,8 @@
 			href="#"
 			title="Build Banner"
 			@click="onCompileBanner">
-				<IconBuild fill="#141414" />
+				<IconBuild fill="#141414"/>
+				<LoadingSpinner :small="true" :loading="isCompiling"/>
 			</a>
 			<a class="banner-actions-icon"
 			:class="{ 'uncompiled': !isCompiled }"
@@ -29,6 +30,7 @@
 			:data-tooltip="bannerCopyTooltip"
 			@click="onCopyBannerToClipBoard">
 				<IconCopy fill="#141414" />
+				<LoadingSpinner :small="true" :loading="isCopying"/>
 			</a>
 			<a class="banner-actions-icon"
 			target="_blank"
@@ -48,7 +50,8 @@ import IconEdit from './IconEdit.vue';
 import IconCopy from './IconCopy.vue';
 import IconBuild from './IconBuild.vue';
 import { relevantTime } from '../relevant_time';
-import { ref } from 'vue';
+import { Ref, ref } from 'vue';
+import LoadingSpinner from './LoadingSpinner.vue';
 
 interface CompileInfo {
 	[key: string]: {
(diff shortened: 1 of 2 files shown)


dashboard/components/BannerActions.vue
```diff
@@ -72,12 +75,15 @@ const WPDE_GITHUB_REPO = 'https://github.com/wmde/wikipedia.de-banners/blob/mast
 
 let editLink: string = props.isWPDE ? WPDE_GITHUB_REPO : CENTRAL_NOTICE_EDIT_URL.replace( '{{banner}}', bannerPageName.value );
 const isCompiled: boolean = !!props.compileInfo;
-
+let isCompiling: boolean = false;
+let isCopying: boolean = false;
```
moiikana, line 79: I think for this to work when you change those values in the methods later you have to use Refs here, like this:
```suggestion
let isCompiling: Ref<boolean> = ref( false );
let isCopying: Ref<boolean> = ref( false );
```
in the methods you can then change the values with
`isCompiling.value = true`;

Correct me if I was testing this wrong, but this code is only working for me when using refs currently. Else, the template values do not get updated and stay `false` (because of how the new "script setup" works)

dashboard/components/LoadingSpinner.vue
```diff
@@ -0,0 +1,73 @@
...
+		<div class="loading-spinner-rect3"></div>
+		<div class="loading-spinner-rect4"></div>
+		<div class="loading-spinner-rect5"></div>
+	</div>
+</template>
+
+<script setup lang="ts">
+import { defineProps } from 'vue';
+
+defineProps( {
```
moiikana, line 14: with the new script setup syntax it might be a good idea for the future to define the props with a const like this:
```suggestion
const props = defineProps( {
```
This way we can use `props.loading` in the code/template and distinguish better between props and local variables. Might improve readability of the code.

In this particular case the `LoadingSpinner` component is a very small component that doesn't do much so I think this is perfectly fine and you don't have to change it, I just wanted to mention this thought.

dashboard/components/LoadingSpinner.vue
```diff
@@ -0,0 +1,73 @@
...
+		<div class="loading-spinner-rect1"></div>
+		<div class="loading-spinner-rect2"></div>
+		<div class="loading-spinner-rect3"></div>
+		<div class="loading-spinner-rect4"></div>
+		<div class="loading-spinner-rect5"></div>
+	</div>
+</template>
+
+<script setup lang="ts">
+import { defineProps } from 'vue';
```
moiikana, line 12: you can remove this line, because `defineProps` should be part of the vue runtime core now.

dashboard/components/BannerActions.vue
```diff
@@ -89,8 +95,10 @@ function bannerCopyHandler( e: Event ) {
 	e.preventDefault();
 	const bannerFileName = `/compiled-banners/${ bannerPageName.value }.js.wikitext`;
 	// TODO start spinner state for copy
```
moiikana, line 97: remember to remove all #todos that you solved

moiikana (approved):

PR #29, opened 2023-03-24
Add Progress Bar

Abban:
This adds the progress bar and the Gotan theme, now renamed to Treedip. It also converts the styles from postcss to scss and uses scss conventions.

Ticket: https://phabricator.wikimedia.org/T332160


src/themes/treedip/ProgressBar/ProgressBar.scss
```diff
@@ -0,0 +1,99 @@
+@use '../theme-variables' as v;
+
+$progress-bar-height: v.$progress-bar-height !default;
```
Abban, line 3: Please don't do this, as it's now a fallback the variable can be moved from the theme-variables into the progress bar stylesheet.

gbirke (approved):

PR #45, opened 2023-04-12
Add Use of Funds

Abban:
This ports the components for Use of Funds over from the application frontend and the styles from the old banner repo.

* Create new accordion and use it on desktop and mobile
* Add check for an existing provisional message from content
* Add tests

Ticket: https://phabricator.wikimedia.org/T332169


src/components/UseOfFunds/FundsDistributionAccordion.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+	<div class="funds-distribution-accordion">
+		<div v-for="( fundsItem, idx ) in applicationOfFundsData" :key="idx" :class="[
+			'funds-distribution-accordion-item',
+			'funds-distribution-accordion-item-' + fundsItem.id,
+			{ 'active' : currentActiveItems.includes( String( idx ) ) }
+		]">
+			<button class="funds-distribution-accordion-item-title" @click="toggleActive( String( idx ) )">
```
Abban, line 8: Typescript throws issues on compile when you try use idx as a number, but is fine when we cast it to a string. Guessing it gets changed somewhere along the build pipeline.

src/domain/Locales.ts
```diff
@@ -0,0 +1,4 @@
+export enum Locales {
```
gbirke, line 1: I think this might be a leftover that can be deleted

test/components/UseOfFunds/CompanyBudgets.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+	{
+		budget: 10,
+		budgetCitation: '',
+		budgetString: '',
+		name: ''
+	}
+];
+
+describe( 'CompanyBudgets.vue', () => {
+	it( 'compares company budgets against the highest one', () => {
```
gbirke, line 28: Thanks for testing this!

gbirke (approved):

src/components/UseOfFunds/FundsDistributionAccordion.vue
Abban: Yeah, but somewhere along the pipeline it must get transposed to a string.

src/domain/Locales.ts
Abban: Yep leftover, but we can use it for setting the locale in the action of the English banners so lets leave it for now

src/components/UseOfFunds/FundsContent.vue
```diff
@@ -0,0 +1,82 @@
...
+				<div>
+					<p v-for="( paragraph, idx ) in orgChartParagraphs" :key="idx" v-html="paragraph"></p>
+				</div>
+			</div>
+			<div class="use-of-funds-orgchart-image">
+				<img :src="content.orgchart.imageUrl"  :alt="content.orgchart.headline"/>
+			</div>
+		</div>
+		<div class="banner_model-section use-of-funds-section-call-to-action">
+			<button class="use-of-funds-button" onclick="location.href='/'">{{ content.callToAction }}</button>
```
gbirke, line 56: This has the wrong handler (`onclick` vs `@click`) and should close the UoF modal instead of going home.

gbirke (changes requested):
Rebased to current main

moiikana (approved):

PR #46, opened 2023-04-13
Split translations into component folders

Sperling-0:


Abban (approved):
LGTM. Just a couple of comments:

It seems like we're spreading them out over the project by putting some in with the dynamic campaign text generator, but I can't think of a better place for them so 👍. If it becomes confusing later on we can try come up with another solution.

We have folders called messages, then files in the banners called translations. Maybe we should rename the banner files to messages.ts, and the variables in Translator for consistency? Then the translator would then get and use messages.

PR #58, opened 2023-04-20
Hook up Skins for Wikipedia.org

Abban:
This makes the banners work on Vector 2022. It also removed Monobook as it's been a long time since it was a default skin.


Abban:
I'll remove the extra config item before merging.

moiikana (approved):

PR #61, opened 2023-04-21
Move banner structure definition to slots

Abban:
The banner component structure is now defined in the root(ish) components.

### DIFF
--- a/banners/desktop/components/BannerCtrl.vue
+++ b/banners/desktop/components/BannerCtrl.vue
@@ -2,18 +2,51 @@
 	<div class="wmde-banner-wrapper" :class="contentState">
 		<BannerMain
 			@close="onCloseMain"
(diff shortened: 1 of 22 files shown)


banners/desktop/components/BannerCtrl.vue
```diff
@@ -2,18 +2,51 @@
...
-			:form-controller="formController"
-			:forms="forms"
-		/>
+			:bannerState="bannerState"
+		>
+			<template #banner-text>
+				<BannerText/>
+			</template>
+
+			<template #banner-slides="{ play }: any">
```
Abban, line 12: We might need to figure out a good way to type these as soemthingother than any

gbirke (approved):
Love it!
Great that you figured out how to pass the event handlers

PR #70, opened 2023-04-25
Add mobile English banner

Abban:
* Copy the Mobile English banner test from the old repo.
* Add new theme.
* Tweak mobile banners

Ticket: https://phabricator.wikimedia.org/T327631


src/components/Slider/KeenSlider.vue
```diff
@@ -77,7 +77,7 @@ const currentSlide = ref<number>( 0 );
 const [ container, slider ] = useKeenSlider( {
 	...props.sliderOptions,
 	initial: 0,
-	loop: true,
+	loop: false,
```
gbirke, line 80: Attention, you now need to set the property to 'true' in all `<BannerSlider>` components in all channes. Maybe change it back to `true` by default, since there is only 1 other component (MultiStepForm) where we need it to be `false`?

src/utils/FormItemsBuilder/fields/PaymentMethods.ts
```diff
@@ -3,9 +3,9 @@ import { FormItem } from '@src/utils/FormItemsBuilder/FormItem';
 type PaymentMethodsKey = 'DIRECT_DEBIT' | 'BANK_TRANSFER' | 'CREDIT_CARD' | 'PAYPAL' | 'SOFORT';
 
 export const PaymentMethods: Record<PaymentMethodsKey, FormItem> = {
-	DIRECT_DEBIT: { value: 'BEZ', label: 'payment-direct-debit', className: 'payment-BEZ' },
-	BANK_TRANSFER: { value: 'UEB', label: 'payment-bank-transfer', className: 'payment-UEB' },
-	CREDIT_CARD: { value: 'MCP', label: 'payment-credit-card', className: 'payment-MCP' },
-	PAYPAL: { value: 'PPL', label: 'payment-paypal', className: 'payment-PPL' },
-	SOFORT: { value: 'SUB', label: 'payment-sofort', className: 'payment-SUB' }
+	DIRECT_DEBIT: { value: 'BEZ', label: 'payment-direct-debit', className: 'payment-bez' },
```
gbirke, line 6: nice change, this makes the classes more consistent

gbirke (commented):
I haven't user-tested it, but code looks good

src/components/Slider/KeenSlider.vue
Abban: They don't loop in the old banners

gbirke (changes requested):
Things I saw when user testing:

* Close button of followup banner does not close the banner
* Donation target on 1st slide is different than progress bar donation target
* slide navigation dots have a border that has a different color (white) than the slide background (light grey, `#202122`)

Abban:
The navigation dots have been like that for a long time. https://en.m.wikipedia.org/wiki/Main_Page?banner=B22_WMDE_Mobile_EN_Test_02_ctrl

Abban:
>Donation target on 1st slide is different than progress bar donation target

I think maybe we've implemented the `goalDonationSum` incorrectly?

gbirke (approved):

PR #72, opened 2023-04-26
Add iPad banner

Abban:
Ticket: https://phabricator.wikimedia.org/T327632


banners/pad/banner_var.ts
```diff
@@ -0,0 +1,73 @@
+import { createVueApp } from '@src/createVueApp';
+
+import './styles/styles.scss';
+
+import BannerConductor from '@src/components/BannerConductor/BannerConductor.vue';
+import Banner from './components/BannerCtrl.vue';
```
Sperling-0: 1. I would request to use 'BannerCtrl' instead of 'Banner'. So that IDE can auto import it.
2. Here, we need to import 'BannerVar.vue', right?

Abban: We name them `Banner` here because the ctrl/var files often get copied and it's easier to just change the import rather than do a find and replace. Technically we don't really care if it's the ctrl or var banner after the import.

banners/pad/banner_ctrl.ts
```diff
@@ -0,0 +1,73 @@
...
+// Change for EN banners
+import messages from './messages';
+import { Translator } from '@src/Translator';
+import DynamicTextPlugin from '@src/DynamicTextPlugin';
+import { LocalImpressionCount } from '@src/utils/LocalImpressionCount';
+import { Formatters } from '@src/utils/DynamicContent/Formatters';
+import { CurrencyDe } from '@src/utils/DynamicContent/formatters/CurrencyDe';
+import { OrdinalDe } from '@src/utils/DynamicContent/formatters/OrdinalDe';
+import { IntegerDe } from '@src/utils/DynamicContent/formatters/IntegerDe';
+import { DeJSONFundsContentLoader } from '@src/utils/UseOfFunds/DeJSONFundsContentLoader';
```
gbirke: You could use the `LocaleFactoryDe` here

gbirke (approved):

PR #75, opened 2023-04-28
Add Documentation

Abban:
* Add BannerConductor doc
* Add Page doc
* Add DynamicContent doc
* Add multistep donation component and form controller doc


docs/BannerConductor.md
```diff
@@ -0,0 +1,26 @@
...
+```
+
+The states are as follows:
+
+* **Initial** This state does nothing, it is used as a default state when the FSM is instantiated.
+* **Pending** This pre-sets the banner height on the Page and starts the banner display timer.
+* **NotShown** When there is a size issue, or if the user interacts with certain elements on the page while the banner display timer is running the FSM will move to this state at the timer end. It will then mark the banner as not shown and fire tracking events.
+* **Showing** This is the banner transition phase. It is in this state when the banner is animating into the page.
+* **Visible** The banner is now finished transition and is visible to the user on the page. It will update the display counts and trigger the displayed tracking events when entered. _Note: This does not control which part of the banner is visible, that is the responsibility of the banner components._
+* **Hidden** When the user has closed the banner.
```
gbirke: Please add some explanation that the states are represented by classes which contain `enter` and `exit` logic to interact with the page object (link to the page documentation), the tracker timeouts, etc. Maybe also mention the `StateFactory` to get a rough impression of which dependencies the states have

docs/DynamicContent.md
```diff
@@ -0,0 +1,24 @@
...
+## Structure
+
+### Generators
+These are for building the sentences and progress bar items. There is a single generator per dynamic text item.
+
+### Formatters
+These are tools used by the generators for localising numbers depending on if the banners are in German or English.
+
+### Campaign Projection
+The current campaign stats are only uploaded every few days during the campaign. The stats are then projected forward from the last upload time in order to keep the data as current as possible. The CampaignProjection is responsible for doing this projection.
```
gbirke:
```suggestion
The campaign team measures the progress of the campaign and uploads the current measurements every few days. We use these measurements to project the numbers displayed in the dynamic text. The calculation starts at the last upload time up to the point in time where the user sees the banner. The CampaignProjection is responsible for doing the calculations for this projection.
```

docs/Page.md
```diff
@@ -0,0 +1,35 @@
...
+* Tracking events.
+* Passing the calculated banner height from our banner javascript to css.
+
+The BannerConductor is then given one of these implementations to perform actions as needed.
+
+## PageOrg (Used on wikipedia.org)
+This implementation is for interacting with wikipedia.org and is a little complex.
+
+### The mw Object
+Wikipedia.org provides this javascript object on the page that the PageOrg class uses for various things such as:
```
gbirke:
```suggestion
Wikipedia.org provides the global `mw` JavaScript object. For PageOrg we wrap the object in a `MediaWiki` interface and use it for things like:
```

docs/Page.md
```diff
@@ -0,0 +1,35 @@
+# The Page Classes
+
+We display banners on wikipedia.org and wikipedia.de. These sites have different html markup, have a different campaign configuration structure, and have different ways of logging events. We use a Page interface with 2 implementations which ensures our code is compatible with these 2 environments.
```
gbirke:
```suggestion
We display banners on wikipedia.org and wikipedia.de. These sites have different HTML markup, have a different campaign configuration structure, and have different ways of logging events. We use a Page interface with 2 implementations which ensures our code is compatible with these 2 environments.
```

docs/Page.md
```diff
@@ -0,0 +1,35 @@
...
+
+They are used for:
+
+* Mounting the banners in the correct place.
+* Tracking events.
+* Passing the calculated banner height from our banner javascript to css.
+
+The BannerConductor is then given one of these implementations to perform actions as needed.
```
gbirke:
```suggestion
In our entry points we give the BannerConductor one of these implementations to perform actions as needed.
```

gbirke (changes requested):

docs/MultiStepDonationForm-and-FormControllers.md
```diff
@@ -0,0 +1,88 @@
+# Multi Step Donation Form and Form Controllers
+
+Since the 2022 campaign a lot of tests have been focused on multistep forms, each with different flows. These tests are different enough functionally that they would require a new `DonationForm` component for each.
+
+In order to avoid duplicating components per-test we opted to instead create a flexible `MultiStepDonationForm` that is provided with different implementations of a `FormController` which contains the logic to run the user path on the form.
```
gbirke:
```suggestion
In order to avoid duplicating components per-test we opted to instead create a flexible `MultiStepDonationForm` that is provided with different implementations of a `FormController` which contains the logic that determines which form page to show, based on the user interaction with the current form page.
```

docs/MultiStepDonationForm-and-FormControllers.md
```diff
@@ -0,0 +1,88 @@
...
+
+This module consists of the following parts:
+
+## `MultiStepDonationForm.vue`
+
+This initialises and presents the form:
+1. It provides the `FormController` with callbacks that allow the controller to handle the flow of the submissions.
```
gbirke:
```suggestion
1. It connects the individual form pages with the callbacks of the `FormController`  to handle the flow of the submissions.
```

docs/MultiStepDonationForm-and-FormControllers.md
```diff
@@ -0,0 +1,88 @@
...
+2. It starts the form `Slider` at page 1.
```
gbirke:
```suggestion
2. It wraps the form pages in a `Slider` component and starts the slides at page 1.
```

docs/MultiStepDonationForm-and-FormControllers.md
```diff
@@ -0,0 +1,88 @@
...
+There are multiple implementations of this (one per type of form) and it is responsible for handling the entire flow of the submission. Events come from the `FormSteps` (with optional extra data), the `FormController` handles the logic of the event and calls one of the callbacks the `MultiStepDonationForm` provided to continue.
+
+## `FormSteps`
+
+Each form consists of one or more steps. Each step:
+
+* Handles its own validation.
+* Emits a `submit` or `next` or `previous` event with data. These events are sent to the `FormController` and allow it to decide what to do next.
+* Optionally resets state when they are entered or exited. For example if a user hits the back button we reset some forms to their default state, and we expect the user to fill them out again to proceed.
+* Optionally modify the `FormModel` directly. The `FormModel` contains only data that is to be posted to the Fundraising Application. That means that forms that contains fields to set this data can set it directly. Some forms, however are for modifying this data as a side effect. These forms pass the extra data to the `FormController` and it decides what should be changed in the `FormModel`
```
gbirke:
```suggestion
* Optionally modify the `FormModel` directly. The `FormModel` contains only data that is to be posted to the Fundraising Application. That means that forms that contains fields to set this data can set it directly. Some forms, however, are for modifying this data as a side effect of the `submit`, `next` and `previous` actions. These forms pass the extra data to the `FormController` and it decides what should be changed in the `FormModel`
```

docs/MultiStepDonationForm-and-FormControllers.md
```diff
@@ -0,0 +1,88 @@
...
+
+## `SubmitForm`
+
+This is a hidden form. It contains the values from the `FormModel` as hidden fields and is submitted by the `MultiStepDonationForm` when `FormController` invokes the `onSubmit` callback. This data is the POSTed to the Fundraising Application as a standard HTTP POST request. 
+
+## In Summary
+
+* The `MultistepDonationForm` is an event bus.
```
gbirke:
```suggestion
* The `MultistepDonationForm` acts like an event bus and keeps track which page is the current one.
```

docs/MultiStepDonationForm-and-FormControllers.md
```diff
@@ -0,0 +1,88 @@
...
+3. From then on it acts as a bus between the `FormController`, `Slider` and the `FormSteps`.
+
+## `FormModel`
+
+This is a [Vuejs composable](https://vuejs.org/guide/reusability/composables.html) and contains the form state. It is globally available. This contains only the data that will be posted to the Fundraising Application and should not be modified for a single test. 
+
+## `FormController`
+
+There are multiple implementations of this (one per type of form) and it is responsible for handling the entire flow of the submission. Events come from the `FormSteps` (with optional extra data), the `FormController` handles the logic of the event and calls one of the callbacks the `MultiStepDonationForm` provided to continue.
+
```
gbirke: Maybe have a short explanation, what the different events mean, esp. the difference between `submit` (="take my form values and either submit or go to the next page") and `next` (="take my form values and jump to the next page that makes sense, don't submit")

docs/MultiStepDonationForm-and-FormControllers.md
```diff
@@ -0,0 +1,88 @@
...
+    FormController ->> MultiStepDonationForm: calls onSubmit callback
+    MultiStepDonationForm ->> SubmitForm: posts form
+```
+
+## Notes for the future
+
+This module isn't perfect. Depending on the direction of the future campaign tests we should improve some things about it.
+
+* The form user flows are badly designed. For example, we don't know the reasoning why some forms clear their own state on back. This needs to be clarified before a redesign happens.
+* `FormControllers` These are where we put the smelly code. Event handling is done per sub-form using switch statements to branch the logic. This could be improved somewhat.
```
gbirke:
```suggestion
* `FormControllers` These are where we put the smelly code. Event handling is done per sub-form using switch statements to branch the logic. This could be improved somewhat.
* Some of the FormControllers are very similar. We might extract them into shared classes for common sets of form steps.
```

gbirke (commented):
Thanks for the Multi-Page-Form explanations! I've added some suggestions for improvements, but they're just suggestions

gbirke (approved):

PR #85, opened 2023-05-08
Implement tracking architecture

gbirke:
Entry points pass mapping factories to Tracker, Tracker classes convert Domain events to specific tracker events. The list of allowed events determines which events are actually sent.

We'll have to see how we can support sending the dynamic banner size to the tracker, maybe we'll refactor that.


src/tracking/TrackerWPDE.ts
```diff
@@ -0,0 +1,108 @@
...
+	 * into event names.
+	 *
+	 * Since we only have 2 channels on WPDE we decided to share the code.
+	 *
+	 * We should regularly clean up this method and remove unused events!
+	 *
+	 * @param {TrackingEvent} event
+	 * @private
+	 */
+	private getEventNameFromEvent( event: TrackingEvent ): string {
```
Abban, line 88: I'd be inclined to add a property to the events for this. It could be implemented as a get property that does the concatenation

gbirke: But then all the events in WPOrg banners would carry around that unneccessary code, which is why we decided to stick that switch statement in here, where it only affects 2 banners, which need this anyway.

moiikana:
I'd like to propose this additional change here:
https://github.com/wmde/fundraising-banners/pull/103 (outdated)
https://github.com/wmde/fundraising-banners/pull/109 (<-- using a union type)

PR #101, opened 2023-05-16
Fix MultistepDonationForm tests

Abban:
* Fixes the tests in the MultiStepDonationForm
* Fix the sub form tests
* Remove FormController interfaces and replace emit data


gbirke (approved):
This will have some conflicts with the latest commit in the tracking branch, where we extracted the form step tracking into a composable for the forms & fixed the tests as well

PR #102, opened 2023-05-22
Add feature tests for multistep forms

Abban:
* Add tests for the different form flows
* Removes form controller references from banner tests
* Update desktop content feature to split test signatures
* Add workaround to vue test-utils for catching form submits in sub-components

### DIFF
--- a/test/features/forms/MainDonation_UpgradeToYearly_CustomAmount_AddressType.ts
+++ b/test/features/forms/MainDonation_UpgradeToYearly_CustomAmount_AddressType.ts
@@ -1,0 +1,62 @@
+import { VueWrapper } from '@vue/test-utils';
+import {
+	expectMainDonationFormGoesToPageOnSubmit,
+	submitMainDonationForm
+} from '@test/features/forms/subForms/MainDonationForm';
+import {
+	expectUpgradeToYearlyFormGoesToPageOnLinkClick,
+	expectUpgradeToYearlyFormGoesToPageOnSubmit
+} from '@test/features/forms/subForms/UpgradeToYearlyForm';
+import { Intervals } from '@src/utils/FormItemsBuilder/fields/Intervals';
+import { PaymentMethods } from '@src/utils/FormItemsBuilder/fields/PaymentMethods';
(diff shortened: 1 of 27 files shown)


test/features/forms/MainDonation_UpgradeToYearly_CustomAmount_AddressType.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Intervals } from '@src/utils/FormItemsBuilder/fields/Intervals';
+import { PaymentMethods } from '@src/utils/FormItemsBuilder/fields/PaymentMethods';
+import { expectCustomAmountFormGoesToPageOnSubmit } from '@test/features/forms/subForms/CustomAmountForm';
+import { expectAddressTypeFormSubmits } from '@test/features/forms/subForms/AddressTypeForm';
+import { AddressTypes } from '@src/utils/FormItemsBuilder/fields/AddressTypes';
+
+export const donationFormFeatures: Record<string, ( wrapper: VueWrapper<any> ) => Promise<any>> = {
+	expectMainDonationFormGoesToAddressFormWhenSofortIsSelected: ( wrapper: VueWrapper<any> ) => expectMainDonationFormGoesToPageOnSubmit(
+		wrapper,
+		4,
```
gbirke: I find these "magic numbers" confusing - I had to look at the function definition of `expectMainDonationFormGoesToPageOnSubmit` to see that the number is a page number, I have to guess that they start at 1, etc. I suggest to create constants in the file.

gbirke (approved):
I like these integration tests - they'll run faster than the Dandy tests and will make sure that the permutations of the form flow are as expected, while keeping duplication to a minimum.

We'll have to see how many form permutations we have in `test/features/forms`. My suggestion would be as follows:
1. Have one file for a combination of *components*
2. If you need a new method for a transition/state change for a specific *step controller*, put it in the file.

We'll also need to talk in the team if we want to archive the tests together with the banners. If we do that (and I'd suggest we do), we'll need to figure out a folder structure that
1. Allows for easy moving of files without adaptation of input paths
2. Allows us to run old tests 
3. But only runs tests on demand / in CI to avoid test times becoming longer and longer

PR #103, opened 2023-05-22
[Followup for Tracking] Add TrackingFeatures enum

moiikana:
- to have more structured data than using strings (at a lot of places)

- this has the drawback of importing the whole Enum at places where just 1-2 strings were needed before

### DIFF
--- a/banners/desktop/components/BannerCtrl.vue
+++ b/banners/desktop/components/BannerCtrl.vue
@@ -49,9 +49,9 @@
 
 		<SoftClose
 			v-if="contentState === ContentStates.SoftClosing"
-			@close="() => onClose( 'SoftClose', CloseChoices.Close )"
-			@maybe-later="() => onClose( 'SoftClose', CloseChoices.MaybeLater )"
-			@time-out-close="() => onClose( 'SoftClose', CloseChoices.TimeOut )"
+			@close="() => onClose( TrackingFeatures.SoftClose, CloseChoices.Close )"
+			@maybe-later="() => onClose( TrackingFeatures.SoftClose, CloseChoices.MaybeLater )"
+			@time-out-close="() => onClose( TrackingFeatures.SoftClose, CloseChoices.TimeOut )"
 		/>
 
 		<FundsModal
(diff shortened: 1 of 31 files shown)


Abban (approved):
Looks good to me, but maybe wait for an opinion from @gbirke as it'll slightly increase our banner sizes.

I would call the enum `Features` as there might be some use for it outside tracking.

Some of the events still have hardcoded strings as their feature, (eg the `ClickAlreadyDonatedEvent`) can you swap those to the enum there before merging?

The Form components also send strings to the `useFormStepShownEvent` composable which could be converted.

gbirke (commented):
I'd like to elaborate a bit on the pros and cons you presented:

- It turns "magic" strings with potential for misspellings into semantic, meaningful types, which allows for autocompletion and "code intelligence" in the IDE
- It follows the "DRY" (Don't-Repeat-Yourself) principle, concentrating the knowledge about the possible values in one place, making it possible to navigate in the IDE ("Show usages" instead of doing a 

Both of these "features" improve two aspects of software quality:
- developer experience 
- code quality (as metric or set of best practices/rules)

But in this case I'd argue that they worsen three other software quality aspects enough that we should really look/discuss if we want to merge this:
- Performance/Sustainability: We're adding about 500 bytes per banner, which will become more in the future as we add more features. I think "keep the banner as small as possible" was one of our main design principles.
- Maintainability/SOLID rules: We're violating the "Single Responsibility" and the "Open for extension, closed for modification" principle: Whenever we create  a new feature, we'll have to edit `TrackingFeatures`
- Architecture: Tracking, in the context of our banners, is a [Cross-cutting-concern](https://en.wikipedia.org/wiki/Cross-cutting_concern) (we track in different layers and modules). I can't decide if making that cross-cutting concern visible is good or a bad thing, but it jsut does not "feel" right to me.

Do you have developer use cases (e.g. new banner, changing a feature, adding a feature) in mind where the centralization would really help us or prevent errors? Or is this change more of a "code quality" and "best practice" change? If it's the latter, then I'd like to skip it due to the tradeoffs.

As a compromise I'd be okay with a union type of the different strings, because the transpiler will drop all that information - we'd have some of the development- and compile-time benefits of an enum (no "show usage", but at least typo protection and a central lookup place with possibility for explanatory comments) and the runtime behavior of plain strings.

moiikana:
closing because using union types seems like a more fitting approach: https://github.com/wmde/fundraising-banners/pull/109/files

PR #109, opened 2023-05-24
Use Union Type for Features in Tracking

moiikana:
Add Union Type for more control over feature name strings, better overview, preventing misspelling etc.

This is based on the discussions in https://github.com/wmde/fundraising-banners/pull/103

### DIFF
--- a/src/tracking/TrackingEvent.ts
+++ b/src/tracking/TrackingEvent.ts
@@ -1,3 +1,20 @@
+export type TrackingFeatureName = '' |
+	'SoftClose' |
+	'Page' |
+	'AlreadyDonated' |
+	'CustomAmount' |
+	'BannerConductor' |
+
+	'MainBanner' |
+	'MiniBanner' |
+	'FullPageBanner' |
+	'AlreadyDonatedModal' |
+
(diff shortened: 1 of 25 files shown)


moiikana:
as far as I understood from our chat "MainBanner" will stay in the union type despite the vue component being called "BannerMain" for simplicity and having the names more consistent in there,
so I won't change anything for now

src/tracking/TrackingEvent.ts
```diff
@@ -1,3 +1,20 @@
+export type TrackingFeatureName = '' |
+	'SoftClose' |
+	'Page' |
+	'AlreadyDonated' |
+	'CustomAmount' |
+	'BannerConductor' |
```
gbirke: I'd remove `BannerConductor` because it's the name of an internal component, that could be confusing for the campaigns team. Either leave it empty or use `Page`

gbirke (approved):
If you're planning to do a merge commit, please remove the `[Followup PR for Tracking]` in the description, otherwise it will show up in the commit history. Or do a "Rebase and merge" commit to avoid the merge commit.

PR #110, opened 2023-05-25
Change app root to Banner components

Abban:


Abban:
Closing. This structure makes what is in each banner very explicit, however it's at the expense of making events confusing and Banners harder to test.

I don't like passing the Banners into the BannerConductor as props, so I'll explore other ways to improve this.

PR #119, opened 2023-05-30
Add fixes for banner size issues and Update Dashboard

Abban:
* Fixed size issue checker not firing
* Fixed name of size issue event


moiikana:
maybe it's possible to add a 
```
z-index: 2;
```
to `.header, .footer` in `dashboard/styles.scss`, to prevent the navigation header and footer from being in the background during scrolling

moiikana (approved):

PR #121, opened 2023-06-05
Move filtered campaigns to script

Abban:


moiikana:
I have to do the following change in line 33 in the same file, then both my IDE and the compiler are happy...

```
:style="{ '--index': parseInt( String(index) ) + 1 }"
```

Abban:
See if this works for you `:style="{ '--index': Number( index ) + 1 }"`

moiikana (approved):

PR #138, opened 2023-06-30
C23_WMDE_Mobile_DE_01

moiikana:
- first pre-campaign test for German Mobile DE

- based on last year's mobile test 14 ctrl
- changes in slides and text in both ctrl and var
- progress bar is removed in both banners

- VAR: less payment options in the form

https://phabricator.wikimedia.org/T340622


banners/mobile/components/FullPageBanner.vue
```diff
@@ -6,7 +6,6 @@
 			</button>
 			<div class="wmde-banner-full-info">
 				<slot name="banner-text"/>
-				<slot name="progress"/>
```
gbirke, line 9: You could leave the slot in for adding it back later more easily, but removing it here probably saves a few bytes ...

gbirke (approved):
The gap in the CTRL banner (caused by the missing credit card) doesn't look nice, but the code does :)

PR #155, opened 2023-07-19
C23_WMDE_Desktop_DE_02

Sperling-0:
- Uses CTRL from test 01
- ctrl has a text change
- VAR is a text and date test

Ticket: https://phabricator.wikimedia.org/T341454


gbirke (approved):
I have tested this and it works, thank you!
If you like, you could amend the 3rd commit and rename the var content components to `BannerSlidesVar` and `BannerTextVar`, so the file casing (PascalCase vs. snake_case) doesn't get mixed up

banners/desktop/content/BannerSlides.vue
```diff
@@ -17,8 +17,8 @@
 	</KeenSliderSlide>
 	<KeenSliderSlide :is-current="currentSlide === 2">
 		<p>
-			Die durchschnittliche Spende beträgt 22,25&nbsp;€, doch bereits eine Spende von 5&nbsp;€ hilft uns weiter. Hat
```
Abban, line 20: We want to keep the `&nbsp;` here and elsewhere to make sure it never line breaks the amount and euro symbol apart. eg:

```
Die durchschnittliche Spende beträgt 22,25
€, doch bereits 5 € helfen uns weiter. Hat
Wikipedia Ihnen in
```

Abban (approved):

moiikana:
remember to create a release of this when merging :information_source:

PR #161, opened 2023-07-25
C23_WMDE_Mobile_DE_02

Sperling-0:
- Uses CTRL from test 01
- CTRL has a text change and close button fix for full page banner
- VAR is a text change, headline change and close button fix for full page banner
- 'banner closed' cookie is not set when closing the fullpage banner for CTRL and VAR

Ticket: https://phabricator.wikimedia.org/T342204


Abban:
@Sperling-0 re "The 'banner closed' cookie shouldn't be set when closing the large modal banner."

If you look in the `BannerCtrl.vue` and `BannerVar.vue` you can see the `FullPageBanner` close event is firing with `CloseChoices.Close`. This can be fired with `CloseChoices.MaybeLater` and it'll then not set the banner closed cookie.

Abban (changes requested):
Please add the feature tests for the VAR banner

PR #162, opened 2023-07-28
Move banner close event tests

Abban:
Close event tests in banners that use the Soft Close feature were located in the Soft Close feature tests.

This moves them to more appropriate features and makes them more explicit.

### DIFF
--- a/banners/mobile/components/BannerCtrl.vue
+++ b/banners/mobile/components/BannerCtrl.vue
@@ -17,7 +17,7 @@
 
 		<FullPageBanner
 			@showFundsModal="isFundsModalVisible = true"
-			@close="() => onClose( 'FullPageBanner', CloseChoices.MaybeLater )"
+			@close="() => onClose( 'FullPageBanner', CloseChoices.Hide )"
 		>
 			<template #banner-text>
 				<BannerText/>
--- a/banners/mobile/components/BannerCtrl.vue
+++ b/banners/mobile/components/BannerCtrl.vue
@@ -17,7 +17,7 @@
 
 		<FullPageBanner
 			@showFundsModal="isFundsModalVisible = true"
-			@close="() => onClose( 'FullPageBanner', CloseChoices.Close )"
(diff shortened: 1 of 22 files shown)


banners/mobile/components/BannerCtrl.vue
```diff
@@ -17,7 +17,7 @@
 
 		<FullPageBanner
 			@showFundsModal="isFundsModalVisible = true"
-			@close="() => onClose( 'FullPageBanner', CloseChoices.Close )"
+			@close="() => onClose( 'FullPageBanner', CloseChoices.MaybeLater )"
```
gbirke: I think these calls should stay a `Close` choice. The `CloseChoices` are for specifying different user choices for a close event - if those choices are available. You can only close the full page banner. While this makes no difference for the legacy tracking (see `event_map.ts`), I think I'd find the term "MaybeLater" confusing if it showed up in a close event log related to the FullPageBanner, because the Banner doesn't have a "maybe later" feature (like our mini nag banner), but only on choice, a close button.

I do get why you changed this - because the *effect* of closing the full page banner acts like a "maybe later" (no setting of a cookie). But for "semantic correctness" of the event logging (the user did not choose between options, but had only the choice to close), I like "Close" better than "MaybeLater", even if that means checks in different parts of the code that examine the `Feature` part of the close event to determine the behavior of sending a tracking or not (if that is the intent of this PR).

Abban: It now emits a Hide choice instead of MaybeLater

src/page/PageWPORG.ts
```diff
@@ -109,6 +109,7 @@ class PageWPORG implements Page {
 			case CloseChoices.NoMoreBannersForCampaign:
 				this._mediaWiki.preventBannerDisplayUntilEndOfCampaign();
 				break;
+			case CloseChoices.Hide:
```
gbirke, line 112: One of the rare cases where `case` fallthrough makes sense

gbirke (approved):

PR #163, opened 2023-07-28
Style mobile full page close buttons

Abban:
The mobile full pages are now modals meaning the close button should be absolutely positioned.

This also adds a background to the buttons so they look okay when they cover other elements.


gbirke (approved):

gbirke:
Code looks good, user-testing as well. Thank you!

PR #171, opened 2023-08-07
Simplify UoF org highlighting

Abban:
Instead of looping and searching for text it now replaces custom tags with spans.

Ticket: https://phabricator.wikimedia.org/T342842


moiikana:
There are 6 files in this PR with changes that have nothing to do with the highlighting, is that just a missing rebase?

moiikana (changes requested):
there are 2 commits from previous days that probably don't belong in this PR and should not get merged, maybe you can remove them with `git rebase interactive` and drop the commits.

the `Simplify UoF org highlighting` commit with the 5 changed files would be approved from my side

moiikana (approved):

PR #228, opened 2023-10-06
C23_WMDE_Desktop_DE_07

Abban:
Var banner has a slightly different form design
and the red border now wraps everything.

Ticket: https://phabricator.wikimedia.org/T346881


src/themes/Treedip/DonationForm/SubComponents/SelectCustomAmountRadioBubble.scss
```diff
@@ -0,0 +1,105 @@
...
+.wmde-banner {
+	@include SelectCustomAmount.layout;
+
+	/*
+	The layout settings of this element should be similar to wmde-banner-select-group-option
+	because the custom amount is presented as an "option" inside the amount select group
+	In the future, we might use tools of the CSS framework (e.g. SCSS mixins)
+	to keep the styles in sync
+	 */
+	&-select-custom-amount {
```
gbirke: If you compare the hierarchy of elements in the amount selection (option elements vs custom amount element), you can see that the hierarchy is different. This class is the equivalent to the `label` element of the regular [SelectGroupRadioBubble](https://github.com/wmde/fundraising-banners/pull/228/files#diff-f5284e8cf5d1a22a5e6e669a7858c4b9e90c528685806b00dd4756114b111c82R20-R47). My approach here would be to try and apply the classes from the `label` element here (and the equivalent subclasses).
If that doesn't work correctly due to flexbox issues (in regular options the `.wmde-banner-option` is a container that has a flex attribute and determines the size of the children and that's missing here), then I'd recommend adding a div with the `wmde-banner-option` around the custom amount component.

src/themes/Treedip/DonationForm/SubComponents/SelectCustomAmountRadioBubble.scss
```diff
@@ -0,0 +1,105 @@
...
+
+			.wmde-banner-select-custom-amount-input {
+				border: 0 none;
+				outline: 0 none;
+				border-radius: 0;
+			}
+		}
+
+		&-input-container {
+			border: 1px solid colors.$gray-light;
```
gbirke: Remove the border here

src/themes/Treedip/DonationForm/SubComponents/SelectGroupRadioBubbles.scss
```diff
@@ -0,0 +1,67 @@
...
+$box-shadow: none !default;
+$background-color: transparent !default;
+$height-label: 21px !default;
+$font-size: 0.095em !default;
+$padding: 0 0 !default;
+$border-label: 2px solid #58575717 !default;
+
+.wmde-banner {
+	&-select-group {
+		&-option label {
```
gbirke: To put a border around the label when its checkbox is selected, you'll need to add a class in the Vue file. For non-bubble components that class won't do anything, so it'll be backwards compatible. Preferrably add the `active` class to the `wmde-banner-select-group-option` (the element that contains the `v-for`).
The selector for that will then be underneath these definitions, something like `&-option.active label` (mind the missing dot before active, it means that the element must have *both* classes)

The reason why we have to do this in the Vue file and not in CSS is the logic of hierarchy in CSS selectors. We always say "I want to style an element that has the following parents", e.g. styling a `label` that's a child of a parent that has the `active` class: `.active label` 
Up until recently, it has been impossible to express a rule for a parent element that depends on some class or pseudoclass of a child. If we had something like a "parent" selector, let's pretend it would be "<", the you could express something like "styling a label element that has a checked checkbox" with a selector like `label < input:checked`. See this StackOverflow question form 2009: https://stackoverflow.com/questions/1014861/is-there-a-css-parent-selector
With the upcoming [:has selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) styling parents based on children will be possible, but that selector is quite new and not supported in Firefox. The selector would then look like `label:has(input:checked)`

moiikana:
rebased main

moiikana (approved):

PR #241, opened 2023-10-20
Fix runtime issues with dev dashboard

Abban:
This stops the intermittent issue we have at runtime where the campaign config sometimes isn't available when we try to filter it in a computed ref.

It also fixes an issue that was throwing warnings when the compileInfo for a banner is missing.


gbirke (commented):
This needs to be rebased so only the fix commit is in this PR

moiikana (approved):

PR #255, opened 2023-10-26
C23_WMDE_Desktop_DE_08

Abban:
- Banner is based on CTRL of last test.
- VAR banner only asks to upgrade when donation amount is under €25
- Adds a new AnimatedText component so we will no longer have to
  add if statements to check for the content of the visitorsVsDonors
  sentence.
- Add new step controller and test for main donation form that
  checks a max amount
- Add new button and test to display correct label depending on amount
- Add new banner tests for amount functionality

Ticket: https://phabricator.wikimedia.org/T348619


banners/desktop/styles/styles.scss
```diff
@@ -8,6 +8,9 @@
 @use 'src/themes/UseOfFunds/UseOfFunds';
 @use 'src/themes/Treedip/defaults';
 @use 'src/themes/Treedip/ButtonClose/ButtonClose';
+@use 'src/themes/Treedip/ProgressBar/ProgressBar' with (
+	$progress-bar-margin: 0 15px
```
moiikana, line 12: what's the reason for this not going into the theme?

moiikana (approved):

banners/desktop/styles/styles.scss
Abban: The iPad used to use the same theme with a different margin IIRC. It can probably be defaulted to `0 15px` in the theme now.

PR #256, opened 2023-10-27
Fix formatting of currentDay

Abban:
This was being formatted incorrectly for German banners, with the day coming after the month instead of before.

I added the day as a template tag in the messages to fix this.


gbirke (approved):
Could you make the German ordinal formatter more semantic by adding the dot in the format function and removing it from the messages? You might need to adapt the campaign day sentence as well

PR #261, opened 2023-11-03
Hide Soft Close on Last Banner Impression

Abban:
- Add Page method to get max impression count
- Add new rules to merged banners.

Ticket: https://phabricator.wikimedia.org/T349186

### DIFF
--- a/banners/english/banner_ctrl.ts
+++ b/banners/english/banner_ctrl.ts
@@ -32,7 +32,7 @@ const mediaWiki = new WindowMediaWiki();
 const page = new PageWPORG( mediaWiki, ( new SkinFactory( mediaWiki ) ).getSkin(), new WindowSizeIssueChecker( 800 ) );
 const impressionCount = new LocalImpressionCount( page.getTracking().keyword );
 const tracker = new LegacyTrackerWPORG( mediaWiki, page.getTracking().keyword, eventMappings );
-const remainingImpressions = Math.max( page.getMaxBannerImpressions() - impressionCount.overallCountIncremented, 0 );
+const remainingImpressions = Math.max( page.getMaxBannerImpressions( 'english' ) - impressionCount.overallCountIncremented, 0 );
 
 const app = createVueApp( BannerConductor, {
 	page,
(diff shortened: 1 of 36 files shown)


banners/english/components/BannerCtrl.vue
```diff
@@ -116,7 +117,11 @@ watch( contentState, async () => {
 } );
 
 function onCloseMain(): void {
-	contentState.value = ContentStates.SoftClosing;
+	if ( props.remainingImpressions > 0 ) {
```
gbirke, line 120: Seeing this repeated across banners makes my DRY reflexes twitch. Mostly because I'd like to give this behavior a name like `softCloseIfNotLastImpression`, etc. OTOH it would make the code less readable and clear, because this if statement is already quite clear and we'd have to pass `remainingImpressions`, `contentState` *and* `onClose` into that "generalized" function to keep it generalized.

Unless we find a way to share `contentState` (and maybe some change/watch methods) across channels (maybe we only have two versions of this, one for desktop and one for mobile), I'd say leave those `if` statements in. And maybe even leave an eventual extraction of `contentState` to another actual banner ticket.

banners/english/banner_ctrl.ts
```diff
@@ -32,6 +32,7 @@ const mediaWiki = new WindowMediaWiki();
 const page = new PageWPORG( mediaWiki, ( new SkinFactory( mediaWiki ) ).getSkin(), new WindowSizeIssueChecker( 800 ) );
 const impressionCount = new LocalImpressionCount( page.getTracking().keyword );
 const tracker = new LegacyTrackerWPORG( mediaWiki, page.getTracking().keyword, eventMappings );
+const remainingImpressions = Math.max( page.getMaxBannerImpressions( 'english' ) - impressionCount.overallCountIncremented, 0 );
```
gbirke, line 35: You could make the entry points a bit nicer and readable by adding `page.getRemainingImpressions(impressionCount.overallCountIncremented)` and moving the `Math.max` and `getMaxBannerImpressions` to `page`.

gbirke (approved):
Looks good to me, two suggested improvements

banners/english/components/BannerCtrl.vue
Abban: Yeah, my thought process went through the same sequences. I ended up putting it into each banner because even though the functionality is shared it's still something that happens per-banner.

banners/english/banner_ctrl.ts
Abban: I'm reluctant to do that because it would create a soft 2 way dependency between `page` and `impressionCount`

PR #262, opened 2023-11-03
C23_WMDE_Mobile_DE_07

Abban:
- Based on the VAR of 06
- VAR banner hides the full page info when the preselect button is clicked.

Ticket: https://phabricator.wikimedia.org/T350457


Sperling-0 (approved):
The banners behave as expected.

PR #265, opened 2023-11-06
Move timer mocking into banner tests

Abban:
Some of the timer mocking in the Banner tests weren't set up correctly and tests were only working because of their order.

This moves the mocking and resetting into `beforeEach` and `afterEach` methods so every feature test is
set up and cleared properly.

### DIFF
--- a/src/components/Slider/KeenSlider.vue
+++ b/src/components/Slider/KeenSlider.vue
@@ -48,6 +48,7 @@ enum SliderPlayingStates {
 
 interface Props {
 	interval?: number;
+	delay?: number;
 	withNavigation: boolean;
 	sliderOptions?: KeenSliderOptions;
 	play?: boolean;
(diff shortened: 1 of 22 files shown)


src/components/Slider/KeenSlider.vue
```diff
@@ -48,6 +48,7 @@ enum SliderPlayingStates {
 
 interface Props {
 	interval?: number;
+	delay?: number;
```
gbirke: You could be more specific about what's being delayed, e.g. `startDelay`. But I can live with `delay`

src/components/Slider/KeenSlider.vue
```diff
@@ -56,10 +57,12 @@ interface Props {
 const props = withDefaults( defineProps<Props>(), {
 	sliderOptions: () => ( {} ),
 	interval: 5000,
+	delay: 0,
 	play: false
 } );
 
 const sliderPlayingState = ref<SliderPlayingStates>( SliderPlayingStates.PENDING );
+const timeout = ref<number>( 0 );
```
gbirke:
```suggestion
const startAnimationTimeout = ref<number>( 0 );
```

gbirke (approved):

PR #272, opened 2023-11-09
C23_WMDE_Desktop_DE_11

Abban:
- Based off the CTRL banner of test 09.
- VAR is a VERY LOUD BANNER WITH MORE VISUAL DISTINCTION

Ticket: https://phabricator.wikimedia.org/T350856


gbirke (approved):
As far as I can see the only difference is adding a drop shadow and moving the border from text to banner, so I find the LOUD BANNER description a bit misleading (because I'd expect more color changes and lots of uppercase from that description)

Abban:
Oh sorry, that was a sarcastic joke based on the test description. Just having a bit of fun with the requirements.

PR #273, opened 2023-11-10
Improve mapping of NotShownEvent

gbirke:
This maps all NotShownEvent instances that don't have "SizeIssue" as a reason to a WMDELegacyBannerEvent with a trackingRate of 0 (we don't need to track those ATM).

Add a test for mapNotShownEvent since the map now contains conditionals.

NotShownEvent now adds a default reason.

NotShownState now always adds the reason when tracking the NotShownEvent.

Delete mapSizeIssueEvent because it duplicated mapNotShownEvent and was not used anywhere.


Abban:
I'm not super into this fix.

- We have 2 different default event reasons
- We no longer log page interaction or invalid namespace reasons
- We check the not shown reason in a couple of places now

I'd say we're better off making it more type safe:

- `customData` is only currently used by the `notShownEvent` and we can inprove that by:
	- Make `TrackingEvent` take in a type parameter `TrackingEvent<T>` that defines the type of the custom data.
	- Add an interface for the notShownEvent data and define it as `TrackingEvent<NotShownCustomData>`.
	- Define the other events as `TrackingEvent<{}>`.
- Remove the `DEFAULT_REASON` from `NotShownEvent`
- Remove the if statement from the `NotShownState` and always pass the data to the `NotShownEvent` constructor.
- Convert the signature of the mapNotShownEvent constructor to take in `NotShownEvent` as it's object.
	- Optionally add a fallback check to validate `customData.reason` there and default it to not known.

This will make TS throw warnings is we map the not shown event incorrectly.


```js
export interface NotShownCustomData {
	reason: BannerNotShownReasons;
	bannerHeight: number;
	viewportWidth: number;
	viewportHeight: number;
}
```

gbirke:
I've tried adding a type parameter to `TrackingEvent`. It looks a bit funky with all those angle brackets and the TypeScript linter explodes any place we're using a plain `TrackingEvent` because we're kind of breaking the Liskov Substitution Principle.
I will prepare a solution that uses plain old polymorphism instead of type parameters: The tracker implementations (regardless if its the legacy schema or the new one) would have some way of converting the custom data anyway, so I think it'll be OK to have a switch statement on the event type instead of the custom data that then tells TypeScript which additional properties (additional to the `TrackingEvent` ones) an event has and then the implementations can build a custom data array out of those properties.

Abban:
We can actually implement `TrackingEvent<T>` in events that don't use it as `TrackingEvent<void>`. And then when it's passed as a parameter it can also use that as it's type eg `public trackEvent( event: TrackingEvent<void> ): void {}`.

That will also allow us to explicitly mark events that have no customData. Check out this [Follow Up](https://github.com/wmde/fundraising-banners/pull/277)

src/tracking/LegacyEventTracking/mapNotShownEvent.ts
```diff
@@ -1,21 +1,37 @@
...
-		},
-		1
-	);
+export const DEFAULT_UNKNOWN_EVENT = 'not-shown-for-unknown-reasons';
+
+/**
+ * @deprecated Will be removed when the new tracking schema is implemented
+ */
+export function mapNotShownEvent( notShownEvent: TrackingEvent<NotShownCustomData> ): WMDESizeIssueEvent|WMDELegacyBannerEvent {
+	if ( notShownEvent.customData.reason === BannerNotShownReasons.SizeIssue ) {
```
Abban, line 19: We don't need this if statement anymore, and we can now always return a WMDESizeIssueEvent.

What we can do is pass it's tracking ratio like `notShownEvent.customData.reason === BannerNotShownReasons.SizeIssue ? 1 : 0`.

At some point we should work out a way to activate tracking for all 3 not shown reasons.

gbirke: I found it more logical to only track a `SizeIssueEvent` when we do want to track screen sizes. I'd leave it like this for now.

PR #275, opened 2023-11-14
Add PR link to dev-dashboard

Sperling-0:
- Add pull request icon
- Use pull request icon
- Clicking on PR icon opens the corresponding PR in a new browser tab


gbirke (changes requested):
This will do requests for every campaign to the GitHub API whenever the dashboard loads. If you look at the headers of the github api responses you can see that there is a rate limit counting down (limiting the amount of API calls to 60 per hour). Depending on usage, we could run into this time limit.
I would suggest the following changes to address this (and make the feature faster in general):
- Done one request to the list of open pull requests (https://api.github.com/repos/wmde/fundraising-banners/pulls?state=open ) Do that request in `DevDashboard` (the most top-level component)
- When the result comes back, build a lookup list (Map or Object with campaign names as keys (using the `title` property of the PR) and pull request URLs as values. Assign the list to a `ref`, e.g. `prList` when you've finished building it.
- Add a nullable parameter for the PR URL to the `Campaign` component. Pass the parameter in the `DevDashboard` component using the value from the previous step (e.g. `:pull-request-url="prList[campaign]"` or `:pull-request-url="prList[currentCampaign]"` )

dashboard/components/DevDashboard.vue
```diff
@@ -21,6 +21,7 @@
 					class="campaign current-branch"
 					:campaign="currentCampaign"
 					:compile-info="compileInfo[ currentCampaign.banners.ctrl.pageName ]"
+					:pull-request-url="campaignUrlMap[currentCampaign.campaign] as string"
```
gbirke:
```suggestion
					:pull-request-url="campaignUrlMap[currentCampaign.campaign]"
```

The type casting in the template is not necessary, you already defined the type of `campaignUrlMap` (see also line 34)

dashboard/components/DevDashboard.vue
```diff
@@ -60,15 +62,26 @@ const campaignList = computed( (): Campaign[] => Object.values( props.campaigns
...
+onMounted( async () => {
+	try {
+		const bannersResponse = await fetch( '/compiled-banners/' );
+		if ( !bannersResponse.ok ) {
+			return;
+		}
+		const fileList = await bannersResponse.text();
+		compileInfo.value = parseCompileInfo( fileList );
+
+		const pullRequestsResponse = await fetch( `https://api.github.com/repos/wmde/fundraising-banners/pulls?state=open` );
```
gbirke:
```suggestion
		const pullRequestsResponse = await fetch( `https://api.github.com/repos/wmde/fundraising-banners/pulls?state=open,closed` );
```

If we also request closed PRs we can jump to PRs that are already closed. I think that's a neat feature to have, but I'll leave that up to you.

dashboard/components/DevDashboard.vue
```diff
@@ -60,15 +62,26 @@ const campaignList = computed( (): Campaign[] => Object.values( props.campaigns
...
+const campaignUrlMap = ref<Record<string, string>>( {} );
+
+onMounted( async () => {
+	try {
+		const bannersResponse = await fetch( '/compiled-banners/' );
+		if ( !bannersResponse.ok ) {
+			return;
+		}
+		const fileList = await bannersResponse.text();
+		compileInfo.value = parseCompileInfo( fileList );
```
gbirke: While we can expect the `/compiled-banners/` route to work reliably (it's provided by the webpack dev server), I want to point out that the `return` on error in lines 70-72 will prevent the querying of the GitHub API if an error occurs.
Also, the `await` prevents the browser loading *both* URLs at the same time. 
This is no blocker and I'll still approve the code. What you could do instead, to make the two fetch requests independent from each other and parallel would be to move the two code snippets to their own function (that function doesn't need try/catch) and then use `Promise.all` to wait for their completion like this:

```typescript
try {
  await Promise.all(fetchCompileInfo(), fetchPullRequestUrlsFromGitHub());
} catch ( error ) {
  console.error( 'Error:', error.message );
}
```

dashboard/components/BannerCampaign.vue
```diff
@@ -17,6 +17,15 @@
 				>
 					{{ campaign.campaign }}
 				</a>
+				<a
+					class="banner-actions-icon"
```
gbirke, line 30: A suggestion for further improving this: With pull request overview URL `https://github.com/wmde/fundraising-banners/pulls` being the default, you could make `class`, `title` and `data-tooltip` dynamic to indicate if there is a "real" PR or just the overview and adjust the color/tooltip-text of the link. Or remove the link with a `v-if="pullRequestUrl==='https://github.com/wmde/fundraising-banners/pulls'"`. I';; leave the decision up to you.

dashboard/components/DevDashboard.vue
```diff
@@ -60,15 +62,26 @@ const campaignList = computed( (): Campaign[] => Object.values( props.campaigns
...
+
+		const pullRequestsResponse = await fetch( `https://api.github.com/repos/wmde/fundraising-banners/pulls?state=open` );
+		const pullRequests = await pullRequestsResponse.json();
+		campaignList.value.forEach( ( campaign: Campaign ) => {
+			const matchingPr = pullRequests.find( ( pr: { title: string; html_url: string } ) => campaign.campaign === pr.title );
```
gbirke: This is a clever way to "fake-type" a JSON result (which usually is `any`). Runtime validation of such results is still a big hassle in TypeScript, even with specialized libraries, so this pattern of "pretending types" like you're doing here (because we *know* the structure will be right) is the appropriate workaround. :+1:

gbirke (approved):
Thank you for this!
No blockers in my comments, just some praise and suggestions for improvements.

dashboard/components/DevDashboard.vue
Sperling-0: Done.

dashboard/components/DevDashboard.vue
Sperling-0: Done.

dashboard/components/DevDashboard.vue
Sperling-0: Done.

dashboard/components/BannerCampaign.vue
Sperling-0: Done.

dashboard/components/DevDashboard.vue
Sperling-0: Thank you 😊

dashboard/components/DevDashboard.vue
```diff
@@ -57,18 +62,37 @@ const compileInfo = ref<Record<string, CompileInfo>>( {} );
...
-		.then( async res => {
-			if ( !res.ok ) {
-				return;
-			}
-			const fileList = await res.text();
+const campaignUrlMap = ref<Record<string, string>>( {} );
+
+onMounted( async () => {
+	try {
+		const [ fileList, allPullRequests ] = await Promise.all( [
```
gbirke: This is not the usage of `Promise.all` I imagined. The code will still block the *processing* of `fileList` and `allPullRequests`, which means that if GitHub is slow (and the 1 second response time of GitHub *is* comparatively slow to the 200ms response time of the `compiled-banners` route), the pull requests update will block the update of the compile info in the UI. The way I imagined this (and which will be non-blocking, and gives names to the fetch-and-process snippets) looks like this example:

```typescript
const updateFileList = async () => {
   const fileList = await fetch( '/compiled-banners/' ).then( response => response.ok ? response.text() : '' );
   if ( fileList ) {
     compileInfo.value = parseCompileInfo( fileList );
  }
}

const updatePullRequestInfo = async () => {
  // put Github fetch and process code here 
}

onMount( async () => {
  try {
    await Promise.all( updateFileList(), updatePullRequestInfo() );
  }
   catch ( error ) {
	console.error( 'Error:', error.message );
   }

} );

```


Sorry to bother you about this (and this is a non-blocking comment and I still approve)

gbirke (approved):

gbirke (approved):

PR #276, opened 2023-11-14
C23_WMDE_Desktop_DE_13

moiikana:
https://phabricator.wikimedia.org/T351199


src/domain/CloseChoices.ts
```diff
@@ -5,8 +5,12 @@ export enum CloseChoices {
 	Hide = 'hide',
 	/** user clicked maybe later button */
 	MaybeLater = 'maybe-later',
+	/** user clicked maybe later button, banner display will be prevented for 7 days */
+	MaybeLater7Days = 'maybe-later',
```
Abban: You don't need to add new choices here, as they're being mapped for the legacy tracking in https://github.com/wmde/fundraising-banners/blob/main/src/tracking/LegacyEventTracking/mapCloseEvent.ts.

What you can do is have the SoftClose X and the AlreadyDonated maybe later fire `CloseChoices.Close` events, then catch them in that mapper to make them distinguishable.

As they then fire normal close events they'll automatically get the 7 days cookie. You can see that here https://github.com/wmde/fundraising-banners/blob/8e3ae9100b525e26db1b4e304bcf2ffcbb47652b/src/page/PageWPORG.ts#L116 (`preventBannerDisplayForPeriod` adds the 7 day cookie, we should update that name at some point)

src/page/PageWPORG.ts
```diff
@@ -125,6 +125,12 @@ class PageWPORG implements Page {
 			case CloseChoices.MaybeLater:
 				this._mediaWiki.preventBannerDisplayForHours( 6 );
 				break;
+			case CloseChoices.MaybeLater7Days:
```
Abban: You can remove these new branches

Abban (commented):
This is a tricky banner, here's some implementation advice:

### Final banner impression close event
>If the close button is clicked and there is no soft close banner displayed, it is logged as a distinguishable event.

The close event when the soft close will no longer be shown is set to MainBanner which you can see an example of here https://github.com/wmde/fundraising-banners/blob/8e3ae9100b525e26db1b4e304bcf2ffcbb47652b/banners/desktop/components/BannerCtrl.vue#L144. That means you can add a check to this file and adjust the name of the event being fired https://github.com/wmde/fundraising-banners/blob/main/src/tracking/LegacyEventTracking/mapCloseEvent.ts

### New main donation form controller
The Upgrade to Yearly and Upgrade to Monthly features should be on separate form pages. That mean you need to add a new `SubmittableMainDonationForm[X]` step controller, that will look at the values the donor selected and submit or change to the correct next step accordingly.

### Upgrade to Monthly
For the upgrade to monthly form step you should add an new form page and controller to handle it, and add an item to https://github.com/wmde/fundraising-banners/blob/main/src/tracking/LegacyEventTracking/mapFormStepShownEvent.ts which will track when that page is shown.

The ticket doesn't say if the upgrade to monthly form submission should be tracked, but I excpect that it should.

banners/desktop/components/SoftCloseWithXButton.vue
```diff
@@ -0,0 +1,86 @@
...
+
+			<div class="wmde-banner-soft-close-column wmde-banner-soft-close-countdown-text">
+				<div
+					class="wmde-banner-soft-close-countdown-text-wrapper"
+					v-html="$translate( 'soft-close-countdown-text', { seconds: secondsRemaining } )"
+				></div>
+			</div>
+		</div>
+
+		<ButtonClose @close="onXButtonClick"/>
```
Abban: This can call `onCloseClick` and you can remove the `maybeLater7Days` emit and `onXButtonClick` function.

src/components/DonationForm/Forms/UpgradeToMonthlyForm.vue
```diff
@@ -0,0 +1,165 @@
...
+
+const secondPageAmount = computed( (): string => currencyFormatter.euroAmount( numericAmount.value ) );
+const secondPageAmountMonthlyFraction = ref( '' );
+
+useFormStepShownEvent( 'UpgradeToMonthlyForm', tracker, props );
+
+// dynamic read-only access for the template variables (so they don't update when the user updates the amount selection on the second page)
+watch( numericAmount, () => {
+	if ( !props.isCurrent ) {
+		suggestedMonthlyFraction.value = Math.ceil( numericAmount.value / 10 );
```
Abban: I'd prefer if this was checked against a map that's passed as a prop, because we can be sure that the fundraising team will want to test different suggested amounts here. You can structure it like:

```js
interface Suggestion {
    min: number;
    max: number;
    suggestion: number;
}

<UpgradeToMonthlyDonationForm :suggestions="[
	{ min: 10, max: 20, suggestion: 2 },
	{ min: 20, max: 30, suggestion: 3 }
]"/>

suggestion.value = props.suggestions.find( s => numericAmount.value > s.min && numericAmount.value <= s.max ).suggestion;
```

src/tracking/events/UpgradeToMonthlyEvent.ts
```diff
@@ -0,0 +1,14 @@
+import { TrackingEvent, TrackingFeatureName } from '@src/tracking/TrackingEvent';
+
+export class UpgradeToMonthlyEvent implements TrackingEvent {
```
Abban: You need to rebase main and change this signature to

```js
export class UpgradeToMonthlyEvent implements TrackingEvent<void> {
...
public readonly customData: void;
```

banners/desktop/components/BannerVar.vue
```diff
@@ -55,6 +59,7 @@
 			@close="() => onClose( 'SoftClose', CloseChoices.Close )"
 			@maybeLater="() => onClose( 'SoftClose', CloseChoices.MaybeLater )"
 			@timeOutClose="() => onClose( 'SoftClose', CloseChoices.TimeOut )"
+			@maybeLater7Days="() => onClose('SoftClose', CloseChoices.Close)"
```
Abban: You can remove this line.

Abban (changes requested):
We need some tests for the UpgradeToMonthlyForm, and new Banner tests that check for the different events coming from the new SoftClose, and AlreadyDonated features. Plus we need new Donation Form Happy Path tests that make sure the user flow is working as expected.

src/components/DonationForm/Forms/UpgradeToMonthlyForm.vue
moiikana: I can understand the intention of doing it like that, but that would also assume there is no relation between the original values and the suggested numbers and they have to be hardcoded all the time (which is not the case so far).
I'd (for now)  suspect that the suggested values will stay kind of deductible.
Maybe we can wait for 1 more test around this feature and see how they might develop it before choosing to hardcode a map like this.

Abban: Yeah but that'll mean that this form page isn't backwards compatible at that point, editing it will break the archive for this test and we'll have to create a new one. So making it a bit more flexible now is worth it. Plus it's way more explicit.

>there is no relation between the original values and the suggested numbers

There should be no relation between the range and the suggested value. If you look at the WMF banners they don't suggest based on math, and that makes sense because we can then suggest different values for different ranges and not just always 10%.

>choosing to hardcode a map like this

The map shouldn't be hardcoded, it should be passed as a prop in the BannerVar component. It's basically configuration and as a bonus is it'll also make it easier to test.

test/features/forms/MainDonation_UpgradeToMonthlyOrYearly.ts
```diff
@@ -0,0 +1,92 @@
...
+		wrapper,
+		Intervals.YEARLY,
+		PaymentMethods.PAYPAL,
+		'25'
+	),
+	expectMainDonationFormGoesToUpgradeYearlyOnHighAmounts: ( wrapper: VueWrapper<any> ) => expectMainDonationFormGoesToPageOnSubmit( wrapper,
+		Pages.UpgradeToYearly,
+		Intervals.ONCE,
+		PaymentMethods.PAYPAL,
+		'200'
```
Abban, line 53: To fix this test you need to add a €200 option to the form items in `test/banners/formItems.ts`

src/components/DonationForm/Forms/UpgradeToMonthlyForm.vue
```diff
@@ -0,0 +1,177 @@
...
+import { Validity } from '@src/utils/FormModel/Validity';
+import ChevronLeftIcon from '@src/components/Icons/ChevronLeftIcon.vue';
+import { Intervals } from '@src/utils/FormItemsBuilder/fields/Intervals';
+import { Currency } from '@src/utils/DynamicContent/formatters/Currency';
+import { Tracker } from '@src/tracking/Tracker';
+import { UpgradeToMonthlyEvent } from '@src/tracking/events/UpgradeToMonthlyEvent';
+import { useFormStepShownEvent } from '@src/components/DonationForm/Forms/useFormStepShownEvent';
+
+interface AmountSuggestion {
+	lowerIntervalLimit: number,
```
Abban: The naming here is a bit confusing because we have a separate interval field in our form. I would call them `lowerRangeLimit` and `upperRangeLimit`

banners/desktop/components/BannerVar.vue
```diff
@@ -38,6 +38,20 @@
...
+					<template #[FormStepNames.UpgradeToMonthlyFormStep]="{ pageIndex, submit, isCurrent, previous }: any">
+						<UpgradeToMonthlyForm
+							:page-index="pageIndex"
+							@submit="submit"
+							:is-current="isCurrent"
+							@previous="previous"
+							:suggestions="[ {
+								lowerIntervalLimit: 11,
+								upperIntervalLimit: 100,
+								suggestedAmount: Math.ceil( formModel.numericAmount.value / 10 )
```
Abban, line 50: 👍 I like this

Abban (approved):
Looks good to me, you just forgot the step controller tests. I'm guessing you're more than tired of this banner so I made a quick follow up with them.

https://github.com/wmde/fundraising-banners/pull/287

moiikana:
> Looks good to me, you just forgot the step controller tests. I'm guessing you're more than tired of this banner so I made a quick follow up with them.
> 
> #287

thanks a lot

PR #278, opened 2023-11-15
C23_WMDE_Mobile_DE_09

Abban:
- Based on CTRL of test 08
- VAR Soft Close has an already donated button

Ticket: https://phabricator.wikimedia.org/T351297


Abban:
I forgot to add the VAR tests. Will add them in the morning.

src/themes/Mikings/SoftClose/SoftClose.scss
```diff
@@ -46,7 +46,6 @@
 			background-color: colors.$secondary;
 			border-radius: 18px;
 			font-size: 14px;
-			line-height: 40px;
```
gbirke, line 49: Isn't removing these line-heights a BC-breaking change for the archived banners?
I'm not sure what we should do in cases like these.

gbirke (approved):
Works and is live.

src/themes/Mikings/SoftClose/SoftClose.scss
Abban: We had issues that the line height was breaking the text alignment on the buttons, this fixes it. So it'll now also be fixed in the archived banners, which I think is fine. We always have the tags in the repo if we need an exact global state.

PR #281, opened 2023-11-16
[DNM] Add FallbackBannerConductor

Abban:
This banner conductor will show a fallback banner if the width is below a threshold or the main banner has a size issue.

Ticket: https://phabricator.wikimedia.org/T351323

### DIFF
--- a/banners/desktop/banner_ctrl.ts
+++ b/banners/desktop/banner_ctrl.ts
@@ -2,8 +2,9 @@ import { createVueApp } from '@src/createVueApp';
 
 import './styles/styles.scss';
 
-import BannerConductor from '@src/components/BannerConductor/BannerConductor.vue';
+import BannerConductor from '@src/components/BannerConductor/FallbackBannerConductor.vue';
 import Banner from './components/BannerCtrl.vue';
+import FallbackBanner from './components/FallbackBanner.vue';
 import getBannerDelay from '@src/utils/getBannerDelay';
 import { WindowResizeHandler } from '@src/utils/ResizeHandler';
 import PageWPORG from '@src/page/PageWPORG';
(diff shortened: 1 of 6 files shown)


banners/desktop/banner_ctrl.ts
```diff
@@ -45,6 +46,8 @@ const app = createVueApp( BannerConductor, {
 	},
 	resizeHandler: new WindowResizeHandler(),
 	banner: Banner,
+	fallbackBanner: FallbackBanner,
+	minWidthForMainBanner: 800,
```
Abban, line 50: This could probably go into the banner config. Actually we probably should be initialisiing that early in the entry point then using values from it so there's no repetition and the values ae more descriptive when passed into the various constructors and props.

gbirke: Sorry, I don't understand this comment - which "banner config" do you mean? `campaign_info.toml`? I'm not sure if we should put too much in there, I think the entry point is a better place to put such "configuration".

gbirke (approved):
I've user-tested this and looked at the code and I'm happy with it.

banners/desktop/banner_ctrl.ts
Abban: No, there's a bannerConfig prop. It also made me think that we could be defining that separately in the entry points eg:

```js
const bannerConfig = {
    minAllowedWidth: 400,
    minWidthForMainBanner: 800,
    bannerDelay: 7500,
    transitionDuration: 1000
};
```

It'd then put those values together and make it a bit more explicit when they're used eg:

```js
new WindowSizeIssueChecker( 400 )
```

would be 

```js
new WindowSizeIssueChecker( bannerConfig.minAllowedWidth )
```

gbirke: That makes perfect sense and is a good idea, thanks!

Abban:
Closing this as it was merged as part of https://github.com/wmde/fundraising-banners/pull/289

PR #283, opened 2023-11-20
Add optional close icon to Soft Close

Abban:
Some of the banners now require an extra close icon in the Soft Close. This adds it as a global option.


gbirke (approved):
Let's discuss (in the next daily) our heuristics when to create new Components (`SoftCloseWithIcon`) and when to add props.

PR #284, opened 2023-11-20
Add new Dynamic Time calculation

Abban:
This adds a new time display item to be used in the banner content

Ticket: https://phabricator.wikimedia.org/T351623

To acceptance test this you'll need to add a ref in a banner content component that can be updated periodically:

```js
const {
	currentDayName,
	currentDate,
	currentTime
}: DynamicContent = inject<DynamicContent>( 'dynamicCampaignText' );

const time = ref<string>( currentTime() );

setInterval( () => {
	time.value = currentTime();
}, 1000 );
```

### DIFF
--- a/src/utils/DynamicContent/DynamicCampaignText.ts
+++ b/src/utils/DynamicContent/DynamicCampaignText.ts
@@ -34,7 +34,7 @@ export default class DynamicCampaignText implements DynamicContent {
 		this._campaignParameters = campaignParameters;
 		this._impressionCount = impressionCount;
 		this._cache = new Map<string, string>();
-		this.currentDateAndTime = this.currentDateAndTime.bind( this );
+		this.getCurrentDateAndTime = this.getCurrentDateAndTime.bind( this );
 	}
 
 	private getCampaignTimeRange(): TimeRange {
@@ -71,7 +71,7 @@ export default class DynamicCampaignText implements DynamicContent {
 	 * Current time returns time to the minute, and needs to be updated dynamically
 	 * This means we can't cache the return value, so instead manually cache the CurrentTime object
 	 */
-	public currentDateAndTime(): string {
+	public getCurrentDateAndTime(): string {
 		if ( !this._currentDateAndTime ) {
 			this._currentDateAndTime = new CurrentDateAndTime( this._translator, this._formatters.ordinal, this._formatters.time );
 		}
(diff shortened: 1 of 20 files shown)


src/utils/DynamicContent/generators/CurrentTime.ts
```diff
@@ -0,0 +1,23 @@
...
+	private _timeFormatter: Time;
+
+	public constructor( translator: Translator, ordinalFormatter: Ordinal, timeFormatter: Time ) {
+		this._translator = translator;
+		this._ordinalFormatter = ordinalFormatter;
+		this._timeFormatter = timeFormatter;
+	}
+
+	public getText( date: Date ): string {
+		return this._translator.translate( 'date-month-time-' + ( date.getMonth() + 1 ), {
```
gbirke: The name of the class is a bit off from the implementation, which returns current date and time. I'm wondering if this isn't too much duplication and we'd be better (in terms of flexibility and DRY-ness) doing `{{currentDate}}, {{currenTime}}` in the banners instead of `{{ currentTime }}`. Although that would lead to an edge case at midnight if we don't update both date and time.

src/utils/DynamicContent/DynamicContent.ts
```diff
@@ -3,6 +3,7 @@ import { DynamicProgressBarContent } from '@src/utils/DynamicContent/DynamicProg
 export interface DynamicContent {
 	currentDayName: string;
 	currentDate: string;
+	currentTime: () => string;
```
gbirke: Is there a reason why this needs to be method instead of a getter?

src/utils/DynamicContent/DynamicCampaignText.ts
```diff
@@ -32,6 +34,7 @@ export default class DynamicCampaignText implements DynamicContent {
 		this._campaignParameters = campaignParameters;
 		this._impressionCount = impressionCount;
 		this._cache = new Map<string, string>();
+		this.currentTime = this.currentTime.bind( this );
```
gbirke: Why is this needed?

src/utils/DynamicContent/generators/CurrentTime.ts
Abban: I was thinking about that name already, lets do `currentDateAndTime`?

src/utils/DynamicContent/DynamicContent.ts
Abban: When it's a getter js seems to cache the value, so it wasn't updating.

src/utils/DynamicContent/DynamicCampaignText.ts
Abban: Because we're using a class, and the method uses `this` internally, so we need to bind the method to the current scope when it's constructed. Otherwise we'll get 'this is not defined' errors.

src/utils/DynamicContent/DynamicContent.ts
```diff
@@ -3,6 +3,7 @@ import { DynamicProgressBarContent } from '@src/utils/DynamicContent/DynamicProg
 export interface DynamicContent {
 	currentDayName: string;
 	currentDate: string;
+	currentDateAndTime: () => string;
```
gbirke: It would be cool if you could document the difference between properties and getter methods in the interface docbloc. Also maybe call this `getCurrentDateAndTime()`

gbirke (approved):

PR #285, opened 2023-11-20
Environment detection

gbirke:
https://phabricator.wikimedia.org/T351627

### DIFF
--- a/src/utils/RuntimeEnvironment.ts
+++ b/src/utils/RuntimeEnvironment.ts
@@ -1,0 +1,54 @@
+/**
+ * This is a subset of the full DOM Location interface,
+ * containing only the properties we need, for easier testing.
+ */
+export interface LocationSlice {
+	search: string;
+}
+
+export interface RuntimeEnvironment {
+	isInDevMode: boolean;
+	runsInDevEnvironment: boolean;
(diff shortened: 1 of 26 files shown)


gbirke:
The new dev-mode-aware getImpressions is currently only used in the Desktop. I'd like to wait for approval/change requests before I change all the other banners.

Abban:
LGTM

src/utils/RuntimeEnvironment.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export interface RuntimeEnvironment {
+	isInDevMode: boolean;
+	runsInDevEnvironment: boolean;
+}
+
+export class UrlRuntimeEnvironment implements RuntimeEnvironment {
+	private _isInDevMode: boolean;
+	private _runsInDevEnvironment: boolean;
```
moiikana, line 16: what are the different purposes of these two fields?

Abban: One marks the `?devMode` query for previewing the banners on WPORG, the other is the `?devBanner` query item from the Fun Forge

Abban (approved):

PR #286, opened 2023-11-21
Simplify legacy close event mapping

Abban:
Legacy close event action names will now be created using the values in the new close event object.


gbirke (approved):
If the PM is fine with this, I am too, this makes the banner smaller and the code easier.

PR #290, opened 2023-11-22
Add FallbackBannerConductor

Abban:
Add a FallbackBanner to Desktop 14

Ticket: https://phabricator.wikimedia.org/T351323

### DIFF
--- a/banners/desktop/components/FallbackBanner.vue
+++ b/banners/desktop/components/FallbackBanner.vue
@@ -1,0 +1,9 @@
+<template>
+	<div class="wmde-banner-fallback">
+		Fallback Banner.
+	</div>
+</template>
+
+<script setup lang="ts">
+
+</script>
--- a/banners/desktop/components/FallbackBanner.vue
+++ b/banners/desktop/components/FallbackBanner.vue
@@ -1,9 +1,103 @@
 <template>
 	<div class="wmde-banner-fallback">
-		Fallback Banner.
+		<ButtonClose @close="onClose"/>
+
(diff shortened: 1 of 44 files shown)


Abban:
I still need to write tests, and figure out where the best place to do that is.

I'm unsure if the fallback banner should be in it's own component folder. It should probably at least have the theme css split and the layout shared somhow.

banners/desktop/components/FallbackBanner.vue
```diff
@@ -0,0 +1,135 @@
...
+			</KeenSlider>
+
+			<button class="wmde-banner-fallback-button">
+				Ja, ich möchte unterstützen.
+			</button>
+
+			<div class="wmde-banner-fallback-mini-footer">
+				<KeenSliderNavigation
+					:current-slide="slideIndex"
+					:slide-count="slideCount"
```
gbirke: You could add the `@click` event here

gbirke (approved):
Looks good to me. As an architecture improvement for the future that might be more flexible, I could imagine a different "banner size negotiation mechanism": Put all the different banner breakpoint sizes into the main banner, each one with a `v-if="bannerForScreenSize('large')"`. `bannerForScreenSize` is a function that's connected to the banner conductor (exact mechanism still to be determined) The banner conductor then has a state called "sizeNegotiation", where it cycles through a map of screensize names (disabling and enabling the booleans returned by `bannerForScreenSize`), renders the banner and sees if it fits. If it does, it stops cycling. If none fits, we switch to the not shown state.

PR #292, opened 2023-11-24
C23_WMDE_Desktop_DE_15

Abban:
Based on the VAR of test 14.
Both banners contain the fallback banner.
VAR uses the upgrade to yearly button form.

Ticket: https://phabricator.wikimedia.org/T351791


test/banners/desktop/components/BannerVar.spec.ts
```diff
@@ -10,11 +10,11 @@ import { TrackerStub } from '@test/fixtures/TrackerStub';
 import { softCloseFeatures } from '@test/features/SoftCloseDesktop';
 import { useOfFundsFeatures } from '@test/features/UseOfFunds';
 import {
-	bannerContentAnimatedTextFeatures, bannerContentDateAndTimeFeatures,
```
gbirke: I don't get why you're removing the tests for `bannerContentDateAndTimeFeatures` - is it because it's a regular feature of banners now that was tested once?

src/tracking/LegacyTrackerWPORG.ts
```diff
@@ -26,13 +26,22 @@ export class LegacyTrackerWPORG implements Tracker {
 		this._runtimeEnvironment = runtimeEnvironment ?? new UrlRuntimeEnvironment( window.location );
 	}
 
+	private shouldTrackEvent( eventRate: number ): boolean {
```
gbirke, line 29: I like it!

gbirke (approved):
Looks good and works as expected.

The only thing left over from the fallback banner PR is the ShownState parameter - Why did you call it `Page` for the regular banner? To me, `Banner` and `FallbackBanner` would make more sense, but maybe you can explain why you chose `Page`

test/banners/desktop/components/BannerVar.spec.ts
Abban: Whoops, copy and paste error

Abban:
@gbirke The tracking feature name was hardcoded to `Page` so I kept it. https://github.com/wmde/fundraising-banners/blob/8fb5115c2ad5fdac25d8bb8422eef5358535c0e2/src/tracking/events/ShownEvent.ts

But you make a good point we should name it Banner now

PR #293, opened 2023-11-24
Never track events with a rate of 0

Abban:
Tracking these events in devMode is confusing for the acceptance testers.

### DIFF
--- a/test/integration/tracking/LegacyTrackerWPORG.spec.ts
+++ b/test/integration/tracking/LegacyTrackerWPORG.spec.ts
@@ -63,7 +63,20 @@ describe( 'LegacyTrackerWPORG', function () {
 		Math.random = oldRandom;
 	} );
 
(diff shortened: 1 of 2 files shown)


test/integration/tracking/LegacyTrackerWPORG.spec.ts
```diff
@@ -63,11 +63,11 @@ describe( 'LegacyTrackerWPORG', function () {
 		Math.random = oldRandom;
 	} );
 
-	it( 'should always track in "devMode" and ignore tracking rate', () => {
+	it( 'should always track in "devMode" and ignore tracking rate when rate is > 0', () => {
```
Abban, line 66: We probably need a better way to test this now because we can never be fully sure that random is above 0.1

gbirke (approved):

PR #294, opened 2023-11-24
Add Workflow to update content repo

gbirke:


Abban:
Are we opening the possibility of getting spammed with PRs in the banner repo with this?

gbirke:
> Are we opening the possibility of getting spammed with PRs in the banner repo with this?

You're right, that's not a good approach. I'll try something else

PR #295, opened 2023-11-27
Update WPDE Tracker

Abban:
We used to track page views on WPDE with trackContentImpression, but we missed porting it when we converted to Vue. This adds it back.

It also improves how it tests the tracking queue.

Ticket: https://phabricator.wikimedia.org/T351823

### DIFF
--- a/test/integration/tracking/TrackerWPDE.spec.ts
+++ b/test/integration/tracking/TrackerWPDE.spec.ts
@@ -1,4 +1,4 @@
-import { describe, beforeEach, it, expect, vi, test } from 'vitest';
+import { describe, beforeEach, it, expect, vi, test, afterEach } from 'vitest';
 import { TrackerWPDE } from '@src/tracking/TrackerWPDE';
 import { CustomAmountChangedEvent } from '@src/tracking/events/CustomAmountChangedEvent';
 import { TrackingEvent } from '@src/tracking/TrackingEvent';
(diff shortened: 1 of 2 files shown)


test/integration/tracking/TrackerWPDE.spec.ts
```diff
@@ -15,11 +15,27 @@ declare let window: Window;
...
+		vi.useFakeTimers();
 	} );
 
-	it( 'tracks events', () => {
-		window.TestTracker = { trackEvent: vi.fn() };
+	afterEach( () => {
+		vi.useRealTimers();
+	} );
+
+	it( 'tracks shown event usong trackContentImpression', () => {
```
gbirke:
```suggestion
	it( 'tracks shown event using trackContentImpression', () => {
```

test/integration/tracking/TrackerWPDE.spec.ts
```diff
@@ -51,47 +65,55 @@ describe( 'TrackerWPDE', function () {
...
+		const trackingQueue: Function[] = [];
+		window.TestTracker = undefined;
 
 		const tracker = new TrackerWPDE(
 			'TestTracker',
 			'TestBanner05',
 			new Map<string, number>( [ [ 'action-1', 1 ], [ 'action-2', 1 ] ] )
 		);
 
+		Object.defineProperty( tracker, '_tracker', { writable: true, configurable: true, value: tempTracker } );
```
gbirke: This test has a bit more knowledge of the actual implementation than I'm comfortable with, testing implementation details instead of just testing behavior. While reviewing, I was also confused what the tests actually does. I'd suggest the following changes:
- rename `tempTracker` to `oldTracker` or `savedTestTracker` or something else, to make it clear in the "arrange" part that we're storing the default one for later.
- override `_tracker` with another tracker, maybe called "trackerSpy". Making sure that it isn't called supports the thing that's being tested, so I'd keep the `notHaveBeenCalled` check
- Drop checking the queue. We don't need to know how the implementation queues the calls.

gbirke (approved):

test/integration/tracking/TrackerWPDE.spec.ts
Abban: Yeah, I think the tests expose that the actual implementation isn't great. We should be injecting in a `PageTrackerFinder` or something so we have total control when the "real" page tracker appears. It'll have BBCs so I was reluctant to do it. We should put it on our list for next year.

`tempTracker` is for checking that the internal `_tracker` wasn't called befor the "real" one was found so maybe a better name would be `nullTrackerSpy` or `missingTrackerSpy`?

Abban: Take a look at those changes.

gbirke (approved):
Thank you for the changes!

PR #297, opened 2023-11-29
C23_WMDE_Desktop_DE_16

Abban:
- Based on the CTRL of test 15
- VAR has a donor heart thing.

Ticket: https://phabricator.wikimedia.org/T351918


banners/desktop/visitorsVsDailyDonorsSentence.ts
```diff
@@ -0,0 +1,7 @@
+const text = `Über {{millionImpressionsPerDay}} Millionen Mal wird unser Spendenaufruf täglich
+	angezeigt, aber nur {{currentDonorsSoFar}} Menschen haben heute gespendet.`;
+
+export function visitorsVsDailyDonorsSentence( millionImpressionsPerDay: number, currentDonorsSoFar: string ): string {
+	return text.replace( '{{millionImpressionsPerDay}}', millionImpressionsPerDay.toString() )
+		.replace( '{{currentDonorsSoFar}}', currentDonorsSoFar );
```
gbirke, line 6: We could make this a bit shorter and more efficient by inlining `text` as a template string

src/tracking/LegacyEventTracking/mapShownEvent.ts
```diff
@@ -9,5 +9,5 @@ export function mapShownEvent( shownEvent: ShownEvent ): WMDESizeIssueEvent|WMDE
 	}
 
 	// We don't track other "not shown" events, hence the trackingRate of 0
-	return new WMDELegacyBannerEvent( 'untracked-not-shown-event', 0 );
+	return new WMDELegacyBannerEvent( 'untracked-shown-event', 0 );
```
gbirke, line 12: It's a "banner was not shown" event, naming it "shown-event" doesn't make much sense to me

gbirke (approved):

src/tracking/LegacyEventTracking/mapShownEvent.ts
Abban: This event is fired when a banner is shown. The comment wasn't updated after a copy and paste

banners/desktop/visitorsVsDailyDonorsSentence.ts
Abban: I think if we're going to be using it in future tests it should go into the dynamic content, then it can use our translator

PR #298, opened 2023-11-29
Ensure latest content when building

gbirke:
Check the commit id of fundraising-frontend-content in package-lock.json
against the latest commit ID from the repo and stop the build process if
there is a mismatch. This prevents us from accidentally downgrading
use-of-funds content.

Ticket: https://phabricator.wikimedia.org/T351385


package.json
```diff
@@ -17,8 +18,9 @@
     "lint:ts": "vue-tsc --noEmit",
     "lint:css": "stylelint '**/*.{scss,vue}'",
     "ci": "npm-run-all lint test",
+    "check-content-version": "ts-node check-content-version.ts",
     "coverage": "vitest run --coverage",
-    "update-content": "npm update fundraising-frontend-content"
+    "update-content": "npm install fundraising-frontend-content"
```
moiikana: what aspect of the difference to using `npm update` did you consider to be "more reliable" here?

gbirke: Thanks for that question. I dimly remembered that `npm update` does not always update  (see [this StackOverflow answer for the specifics](https://stackoverflow.com/questions/12478679/npm-install-vs-update-whats-the-difference)) and when I ran `npm install fundraising-frontend-conten` it always reliably updated my outdated version. Thus I assumed that `install` is more reliable. I tried to search for our specific case (non-semver, pointing to a branch) but did not find anything useful. I asked ChatGPT for our specific scenario and it told me that `npm update <packagename>` will always update. I tested it on my command line and saw that it really updated. So I'll change that script back to `update`. If we encounter instances where it doesn't update, we can try out if `install` is more reliable.

moiikana (approved):

PR #302, opened 2023-12-04
Fix timer issues on mobile banner tests

Abban:
The mobile banners now have a dynamic time text item that are updated using setInterval. The timers were configured to be stopped in onUnmount, but those components are acutally never un mounted making the timers infinite which caused the soft close tests to break. This change makes them stop if the banner moves on from the state where they're visible.

Also adds tests for the dynamic text in mobile banners and fixes the docblock in the createDonationURL and
createFormActions functions.

Ticket: https://phabricator.wikimedia.org/T352173


moiikana:
rebased parent branch

moiikana (approved):

PR #311, opened 2023-12-07
C23_WMDE_Mobile_DE_12

moiikana:
https://phabricator.wikimedia.org/T352889


gbirke (approved):
The extra style file (`C23_WMDE_Mobile_DE_09` should be incorporated as a `Mikings` variant file, e.g. `SoftClose/SoftCloseWithImprovedButton`) to be named more properly and not after the test which gives no indication what it overrides. But you can do that in the mobile-13 PR

Abban (approved):

PR #314, opened 2023-12-08
Hook up isLateProgress configuration

Abban:
We forgot to port this over from the old banner project. It is now part of the progress bar dynamic content meaning it should be hooked up automatically.


gbirke (approved):

Abban:
This was merged with desktop 20

PR #318, opened 2023-12-12
C23_WMDE_Desktop_DE_21

Sperling-0:
- Based on the VAR of test 20
- Add a composable file which sets address type to anonymous except for the direct debit payment type
- If it's not a single time payment, the main donation form's submit button label is changed depending on the selected payment method, add new component for this
- Submitting the form in the variant banner leads to donation/add and additionally submits the address type "anonymous" when the payment method is NOT direct debit
- Add the tests for added features

Ticket: https://phabricator.wikimedia.org/T352973


gbirke (approved):
Looks good and works as specified. To merge this, I'd create a new branch from main and cherry-pick your 3 commits into that branch. then replace this branch/pr with the new branch

test/components/DonationForm/StepControllers/SubmittableMainDonationFormSinglePageAnonymous.spec..ts
```diff
@@ -0,0 +1,38 @@
...
+		const stepNavigation = { goToStep: vi.fn(), submit: vi.fn() };
+		const controller = createSubmittableMainDonationFormSinglePageAnonymous( formModel );
+
+		await controller.submit( stepNavigation, {} );
+
+		expect( stepNavigation.submit ).toHaveBeenCalledOnce();
+		expect( stepNavigation.submit ).toHaveBeenCalledWith( new BannerSubmitEvent( 'MainDonationForm' ) );
+	} );
+
+	it.skip( 'Leads directly to the page of the external service provider on submit', () => {
```
Abban: You should delete this

Abban (changes requested):
Can you modify the test in `test/components/composables/useFormAction.spec.ts` to also check that the correct action is returned when the paymentType is direct debit?

test/components/DonationForm/StepControllers/SubmittableMainDonationFormSinglePageAnonymous.spec..ts
Sperling-0: Done.

src/components/SubmitButton/VariableSubmitButton.vue
```diff
@@ -0,0 +1,35 @@
...
+const formModel = useFormModel();
+const translator = inject<Translator>( 'translator' );
+const { paymentMethod } = formModel;
+const { interval } = formModel;
+
+const submitButtonLabel = computed( (): string => {
+
+	if ( interval.value === Intervals.ONCE.value ) {
+		return translator.translate( 'submit-label-short' );
+	} else {
```
Abban: Because the first branch in the conditional returns you don't need the `else` so you can reduce nesting by removing the checks for payment method out and getting rid of it.

test/components/composables/useFormAction.spec.ts
```diff
@@ -24,4 +25,24 @@ describe( 'useFormAction', () => {
 
 		expect( formAction.value ).toBe( expectedAction );
 	} );
+
+	test.each( [
```
Abban: If you want you can integrate this test and the previous one, the ones from previous would need to be set to use the PAYPAL payment method.

Abban (changes requested):
Please don't forget to write tests for:

- Form Step Controllers:
	- `SubmittableMainDonationFormAnonymous`
	- `SubmittableUpgradeToYearlyAnonymous`
- `VariableSubmitButton` This shouldn't be tested as part of the BannerVar tests because if CTRL wins then the test for the button will get deleted, but the button won't. So you should make a new test that only mounts the button component and move the test from BannerVar to that.

I think the existing donation form happy path tests in BannerVar cover the flows okay so we don't need any new ones there.

I think the `VariableSubmitButton` component is in the wrong place, we have similar buttons in `DonationForm/Forms` and it should be in there with those. It should also be named more explicitly eg `MainDonationFormButtonVariablePaymentType`.

Abban:
I've just spotted that the tests I requested are already implemented as part of https://github.com/wmde/fundraising-banners/pull/319 so we should look at merging the work from the 2 PRs.

test/components/composables/useFormAction.spec.ts
Sperling-0: Done.

src/components/SubmitButton/VariableSubmitButton.vue
Sperling-0: I made the changes, please check if it is correct.

Abban (approved):

PR #320, opened 2023-12-14
C23_WMDE_Desktop_DE_22

Abban:
- Based on the VAR of test 20
- VAR has a double progress bar

Ticket: https://phabricator.wikimedia.org/T353329


banners/desktop/banner_ctrl.ts
```diff
@@ -59,7 +59,8 @@ app.use( DynamicTextPlugin, {
 	date: new Date(),
 	formatters: localeFactory.getFormatters(),
 	impressionCount,
-	translator
+	translator,
+	urgencyMessageDaysLeft: 45
```
gbirke: Is this a leftover from some experiment?

gbirke (approved):
Code looks good.
I'm wondering how you tested locally, because the local overrides in `src/environment/dev/CampaignParameterOverride.ts` prevented me from previeweing the proper values. Maybe you can comment out or delete most of the overrides there, so the local display uses real campaign values

banners/desktop/banner_ctrl.ts
Abban: No, the `campaignDaySentence` shows a more urgent message if we're within `N` days of the end date. It defaults to 10 but they wanted to activate it early and 45 was the random number I picked that was greater than the current days left when it was activated.

Abban:
> I'm wondering how you tested locally,

I didn't, I uploaded it and did the final test on Central Notice

PR #321, opened 2023-12-14
Set page body fixed on modal open

Abban:
This will prevent scrollbars from interfering with each
other when the user views a full page modal. It will also
stop the page scrolling behind the user when they're looking
at a modal and they scroll their mouse wheel.

Ticket: https://phabricator.wikimedia.org/T353435


moiikana:
TODO:
 before merging, clean up "@Abban
[TEMPORARY COMMIT FOR TESTING](https://github.com/wmde/fundraising-banners/pull/321/commits/1b507fa4d284296b39ec3edbb8939221a22e2cff)"

moiikana (approved):

PR #328, opened 2023-12-19
Add minimisable banner

Abban:
This adds a sample minimisable banner that may be used in a future test.


Abban:
Closing because this is now implemented as https://github.com/wmde/fundraising-banners/pull/333

PR #334, opened 2023-12-22
C23_WMDE_Desktop_DE_25

Sperling-0:
- Based on the VAR of test 24
- VAR has a collapsible banner

Ticket: https://phabricator.wikimedia.org/T353639


gbirke:
This is broken (my fault during a rebase/cherry pick), so we'll recreate it

PR #360, opened 2024-01-17
Add ADR on commit structure

gbirke:


docs/adr/005-git-commit-structure.md
```diff
@@ -0,0 +1,143 @@
...
+    Test_25, while Test_23 is still unmerged), it becomes impossible to
+    merge earlier tests, because they would appear in the wrong order in
+    the succession of merged branches and merging would also resolve the
+    wrong conflicts.
+
+These scenarios have happened several times during the 2023 campaign.
+
+## Decision
+
+### Option 1: Get rid of "current banner" concept
```
Abban: We already point the config file to the banner entrypoints, so having a structure like `banners/desktop/C24_blah_01, banners/desktop/C24_blah_02` would be an easy enough change.

It would also remove the archive step (and the archive), the prepare campaign step would be duplicating and renaming the banner folder for the next test. Way less merge issues because different banner branches would be working on code in different places so not confusing git so much.

The downside would be a lot of folders in the busy banner channels towards the end of the campaign, but I can deal with that. I'd prefer it over having to update a symlink every test.

moiikana: I also like the clear approach of having just straight up "ready to work" folders for all the banners, 
even though I fear the same as Abban -> more mistakes, accidental changes in other folders. 
Would be so cool if you could _lock_ folders locally in your IDE like with layers in some painting software.

docs/adr/005-git-commit-structure.md
```diff
@@ -0,0 +1,162 @@
...
+   (i.e. a previous banner this banner is based on).
+2. `Prepare CTRL for [CAMPAIGN_NAME]`: In this commit we prepare the control banner
+   using the winner of the previous test as it's base. This will allow reviewers to
+   view a diff of the changes made to create the base from the previous test.
+3. `Implement VAR for [CAMPAIGN_NAME]`: We then implement the variant banner.
+
+We can now merge banner feature branches in any order, without conflicts
+in the banner code. The only conflict that can occur is in
+`campaign_info.toml`, but we'll resolve that conflict with the following
+rule: Always merge the version that has the higher test number.
```
Sperling-0: "Always merge the version that has the higher test number."
What is the meaning of _higher test number_ here?

gbirke: I'll add an explanation and example

gbirke: I've added another commit with example

Abban (approved):

PR #372, opened 2024-01-31
Draft a schema file for our banner events

gbirke:
This file will be modified and moved into the official schema repo, this is just a draft and we need to clarify some things

Ticket: https://phabricator.wikimedia.org/T336842


Abban:
The Identifier Naming Rules [here](https://wikitech.wikimedia.org/wiki/Event_Platform/Schemas/Guidelines) say to use snake_case

PR #493, opened 2024-07-16
[FollowUp] Remove !important rule

moiikana:
-  using `!important` in CSS is often a sign of a code smell when writing CSS and we try to avoid it,
    because it means there might was trouble with the CSS's specificity rules (
    see e.g. https://csswizardry.com/2012/11/code-smells-in-css/ -> "Reactive !important")
	`!important` should only be used in very exceptional cases
    
- exchange the p tag (which have strong rather global css rules) with a normal div tag
    (that is easier to style, we don't have to overwrite p rules)


Sperling-0 (approved):

Abban:
Getting rid of the `!important` attribute here is the correct thing to do, but replacing the paragraphs with divs isn't, as it's not semantic. Instead the paragraphs should kept and the margin overidden using higher css specificity.

PR #527, opened 2024-08-28
Add swatch for managing treedip colours

Abban:
This moved the colours from the scss variables into
css variables meaning we can handle dark modes.

- Updated EN/iPad/WPDE banners that use Treedip
- Deleted some unused components

Ticket: https://phabricator.wikimedia.org/T372558


Abban:
I don't have all the components available to test in dark mode, so as they're used in banners we need to adjust the colours

banners/english/C24_WMDE_Desktop_EN_00/styles/styles.scss
```diff
@@ -1,4 +1,11 @@
 // This is the file where we import the theme-specific component styles
+@use 'src/themes/Treedip/swatches/vector-2022' with (
+	$slider: true,
+	$progress-bar: true,
+	$select-group-bubble: true,
+	$upgrade-to-yearly: true,
+	$fallback-banner: true,
+);
```
gbirke, line 8: I'm very torn here. Using parameterized swatches makes sure that we only output what we really need, making the banners compact and performant.  
But the developer experience and structural clarity suffers:

- The central style file has now 2 places to change when adding new components/features. This is not very DRY and has error potential (adding a component and forgetting its swatch or assuming that adding the swatch would style the component). The good thing is that the errors can be seen visually and the solution will be apparent. It's just more work for us.
- The features don't map 1:1 to components
- When we introduce a new feature we have to remember to edit all the swatches in all the themes to keep a consistent interface

But overall it's a good start, does what we want and I'm willing to try this solution to see how much we suffer and long for something else.

gbirke (approved):
I'd love to get some documentation on
- General introduction how theme colors work and where to put them
- Intro to the swatch "interface"
- What to do when adding a new feature/component.

I can see how it works by looking at the code, but an explicit document would be nice.

I' approving, but haven't user-tested yet.

banners/english/C24_WMDE_Desktop_EN_00/styles/styles.scss
Abban: Yeah I thought about putting the variable declarations into each of the component styles. But that would mean we need 4 difference selectors in each one to apply them and lead to a lot of repetition and have a lot of potential for human errors from that. It would also make finding all the variables a little harder as they'd be spread out across the components.

We could drop the flags altogether and include all the variables in all the banners, but because they're defined 4 times I wanted to cut it down to only what we need.

PR #528, opened 2024-08-30
C24_WMDE_Desktop_EN_01

Abban:
- Banners are based on CTRL of C23_WMDE_Desktop_EN_05
- Both banners integrate with Wikipedia's Dark Mode
- VAR banner submits to the contact required address form layout


banners/english/C24_WMDE_Desktop_EN_01/banner_ctrl.ts
```diff
@@ -67,7 +67,7 @@ const currencyFormatter = localeFactory.getCurrencyFormatter();
 
 app.provide( 'currencyFormatter', currencyFormatter );
 app.provide( 'formItems', createFormItems( translator, currencyFormatter.euroAmount.bind( currencyFormatter ) ) );
-app.provide( 'formActions', createFormActions( page.getTracking(), impressionCount, { locale: Locales.EN } ) );
+app.provide( 'formActions', createFormActions( page.getTracking(), impressionCount, { locale: Locales.EN, ap: '0' } ) );
```
Sperling-0, line 70: This change should be part of `Prepare CTRL for C24_WMDE_Desktop_EN_01` commit, right?

Abban: I always think of the prepare CTRL commit as the one that resets the banner code in preparation for the next test. Eg doing only the bit where in the ticket it says "banners are based on control/variant of test blah". So purely creating the baseline for the next test.

Then the changes for the next test are done in the VAR commit. So in this case the VAR should send the donor to ap=1 and the CTRL to ap=0.

This means there's a clear cut-off between the prep and the next test implementation. So for example if the test was to be changed at the last minute we can reset the branch to the prepare CTRL commit and we're ready to make the new var. If we'd made changes to the CTRL in that commit we would also have to remember that and revert them.

Abban: An exceptiopn to that is if the ticket has a section called "Both banners should:" which means that the baseline for the next test is being adjusted between the last test and before we implement the next one. I put those changes into the prepare CTRL commit.

gbirke (approved):

banners/english/C24_WMDE_Desktop_EN_01/banner_ctrl.ts
gbirke: Thank you for this explanation! I did it differently in Desktop 04, where I added the param to both banners in the CTRL commit and then changed its value in the VAR commit, but your explanation also makes sense, esp. for param changes.

PR #529, opened 2024-09-02
C24_WMDE_Mobile_DE_04

Sperling-0:
Ticket: https://phabricator.wikimedia.org/T372555


banners/mobile/C24_WMDE_Mobile_DE_04/styles/FullPageBanner.scss
```diff
@@ -0,0 +1,102 @@
+@use '@src/themes/Mikings/variables/colors';
+@use '@src/themes/Mikings/variables/globals';
+@use '@src/themes/Mikings/variables/breakpoints';
+
+.wmde-banner {
+	--slider-chevron: #{colors.$black-light}; // Temporary fix until mobile banners use swatches
```
Abban, line 6: Oh, when I created this variable I didn't notice that the chevrons are being used elsewhere. So this variable name doesn't make sense in the context of using it as a back icon. I'll think about a better way.

moiikana (approved):

PR #542, opened 2024-09-10
Add document explaining how swatches work

Abban:
The setup for swatching on banners is complicated and this
document is aimed at helping people get their heads around
how it works.

### DIFF
--- a/docs/Swatches.md
+++ b/docs/Swatches.md
@@ -1,0 +1,161 @@
+# Swatches
+
+The Vector 2022 Wikipedia skin now supports dark mode and that means our banners are now required to support it.
+
+## Changes
+
+- Removed the `_colors.scss` file that contained our colour settings.
+- Added a swatch folder into the themes that implement the colours as [css vars](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties).
+- Added swatches to the Use of Funds, and Fallback banner themes.
+
+## Implementation Notes
+
+### Naming Scheme Changes
+
+Because we now need fine-grained control of how each colour for each part of the banners are defined, we needed to change the naming scheme.
+
+Our old structure for colours was based around the names of the colours:
+
+```scss
+$primary: #386d94;
+$secondary: #da9000;
+$error: #dd3333;  
+  
+$white: #ffffff;  
+$black: #000000;  
+$gray: #54595d;  
+$gray-medium: #a2a9b1;  
+$gray-light: #e5e5e5;  
+```
+
+While we still keep variables for the colours for convenience, the new structure gets rid of this convention and defines the colour variables per feature:
+
+[color_light.scss](https://github.com/wmde/fundraising-banners/blob/main/src/themes/Treedip/swatches/light.scss)
+```scss
+$white: #ffffff;  
+$black: #000000;  
+  
+$grey100: #f8f9fa;  
+$grey150: #dddddd;  
+$grey200: #bbbbbb;  
+$grey300: #b7b7b7;  
+$grey500: #808080;  
+$grey600: #747474;  
+$grey700: #202122;  
+  
+$red600: #990a00;  
+  
+$blue600: #4465a7;  
+$blue700: #2a4b8d;
+
+--main-background: #{$white};  
+--main-color: #{$grey700};  
+--main-title-line-background: #{$red600};  
+--main-border-bottom: #{$grey200};  
+--message-border: #{$red600};
+```
+
+This means that rather than defining global colours to be used in our styles we now define css variables on a component/feature level.
+
+### Using Swatches With Skins
+
+Because not all skins support dark mode, and when they do they'll most likely have slightly different ways of implementing it, we need a flexible way of including the vars into the banners that run on them.
+
+This resulted in the swatch files being split up into colour definitions and implementations, for example the files in the swatch folder look like this:
+
+```
+/swatches
+    color_dark.scss // Colours for dark theme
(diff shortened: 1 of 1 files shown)


docs/Swatches.md
```diff
@@ -0,0 +1,113 @@
...
+/swatches
+    dark.scss // Colours for dark theme
+    default.scss // Default implementation for skin that doesn't have dark mode
+    light.scss // Colours for the light theme
+    vector-2022.scss // Vector 2022 implementation that includes dark mode
+```
+
+The [`dark.scss`](https://github.com/wmde/fundraising-banners/blob/main/src/themes/Treedip/swatches/dark.scss) and [`light.scss`](https://github.com/wmde/fundraising-banners/blob/main/src/themes/Treedip/swatches/light.scss) files are the colour definitions. They contain all the variables to control all the colours in the banner. They provide these vars through a mixin.
+
+The [`default.scss`](https://github.com/wmde/fundraising-banners/blob/main/src/themes/Treedip/swatches/default.scss) and [`vector-2022.scss`](https://github.com/wmde/fundraising-banners/blob/main/src/themes/Treedip/swatches/vector-2022.scss) files are the per-skin implementations of the colours. They contain css selectors that target the light or dark mode skin and insert the variables from the colour definitions into them.
```
gbirke: So there are 2 types of swatch files - color definitions (to be included) and theme-related ones. Maybe you can add a `color_` prefix to the color ones to group them and tell them apart? Or use a subfolder?

docs/Swatches.md
```diff
@@ -0,0 +1,113 @@
...
+
+- Removed the `_colors.scss` file that contained our colour settings.
+- Added a swatch folder into the themes that implement the colours as [css vars](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties).
+- Added swatches to the Use of Funds, and Fallback banner themes.
+
+## Implementation Notes
+
+### Naming Scheme Changes
+
+Because we now need fine-grain control of how each colour for each part of the banners are defined we needed to change the naming scheme.
```
gbirke:
```suggestion
Because we now need fine-grained control of how each colour for each part of the banners are defined, we needed to change the naming scheme.
```

Our English teacher yelled at us when we started sentences with "Because", but maybe that's an outdated rule, I trust your native speaker skills here.

docs/Swatches.md
```diff
@@ -0,0 +1,113 @@
...
+2. When the donor has selected *Dark Mode*.
+3. When the donor has selected *Automatic* and their OS is set to *dark mode*.
+4. When the donor has selected *Automatic* and their OS is set to *light mode*.
+
+If/when another skin starts to implement dark mode we will need to create a new implementation for that.
+
+### Optional Variables
+
+Because all banners don't include all components/features, we define [default scss variables](https://sass-lang.com/documentation/at-rules/use/#configuration) as boolean flags in the swatch implementations that allow you to include only the variables for the features the banner uses. This helps keep the compiled css a little smaller, especially on banners that need to declare the variables 4 times.
+
```
gbirke, line 92:
```suggestion
Whenever you add a new component/feature, you'll have to add a section to `colors_light.scss` and `colors_dark.scss` where you define the new color names:

example here

```

docs/Swatches.md
```diff
@@ -0,0 +1,113 @@
...
+### Using Swatches
+
+You import the swatch for the skin the banner is running on at the top of the banner's `style.scss`. Then you pass through flags to include the css vars you need for your features/components eg:
```
gbirke:
```suggestion
You import the swatch for the skin the banner is running on at the top of the banner's `style.scss`. Then add the flags for the features/components used in your banners, e.g.:
```

docs/Swatches.md
```diff
@@ -0,0 +1,113 @@
...
+);
+```
+
+Not all banner themes use swatches yet, but all banners will need to include the Use of Funds swatches:
+
+```scss
+@use 'src/themes/UseOfFunds/swatches/default' as uof-default;
+```
+
+Please note the use of namespacing here as the compiler will complain if we use 2 modules with the same name and default is already defined when importing the main swatch.
```
gbirke, line 145: You could add an example of how to convert a theme to swatches

gbirke (changes requested):

docs/Swatches.md
Abban: I'll have to do this in another PR

Abban: Actually, maybe we should also prefix the skin implementations with skin_ too?

docs/Swatches.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Converting a theme to a swatch:
+
+When you need to convert an existing theme to use the swatches you should follow these steps:
+
+1. Create `color_light.scss` and `skin_default.scss` files in a swatches folder.
+2. Copy the colours from the `variables/_colors.scss` file into the `color_light.scss` and delete `variables/_colors.scss`. Deleting the old file first causes the compiler to throw errors which makes it a little easier to find all the usages.
```
moiikana: Question:
Do we have a strategy on how to create/deduct the colors for a `color_dark.scss` file yet?
Or do we basically just leave that to the designer, providing us with a dark design + colors via Figma?

Abban: This is another question for the PM. I can do the swatch work myself, but I'm in limbo about how much UX availability we have and if I can just go ahead and do it or not.

If I was to do it I'd do it straight in code, no Figma needed.

gbirke (approved):

PR #552, opened 2024-09-20
Add dark mode for Minerva themes

Abban:
Minerva now supports dark mode so that means our mobile themes also need to support it.

This adds a light swatch only to the Minerva/Mikings theme as we need UX input to know what colours the dark mode should be.

There is a bug in our dev environment where clicking the dark mode preview on the mobile DE banner doesn't activate dark mode on Minerva even though the `minervanightmode=1` is present in the URL. This works find on enwiki so I don't know why it doesn't work on dewiki, it might be because dewiki Minerva doesn't yet ofiicially support dark mode.


Abban:
Please give this a good acceptance test as it was tedious work that happened at 4pm on a Friday.

Sperling-0:
Following is the behavior when I user tested it:

Desktop de 5 : No dark theme, normal banner appears for both theme options
Mobile de 4: Dark theme present, but normal banner appear for both theme options
WPDE mobile 0: No dark theme, banner does not load for both theme options

Sperling-0 (commented):
Following is the behavior when I user tested it:

Desktop de 5 : No dark theme, normal banner appears for both theme options
Mobile de 4: Dark theme present, but normal banner appear for both theme options
WPDE mobile 0: No dark theme, banner does not load for both theme options

Abban:
Desktop DE 5: Doesn't have a dark swatch as dewiki is still on Vector
Mobile de 4: Doesn't have a dark swatch as we need UX input for it
WPDE Banners: Won't have a dark theme unless we decide to support dark mode on wikipedia.de

I should have been clearer in the description of what I changed, I'll clarify it.

gbirke (approved):
User-tested it and saw no change, it's looking good (also code-wise)

PR #555, opened 2024-09-26
C24_WMDE_Desktop_DE_06

moiikana:
- based on DE 05 ctrl

https://phabricator.wikimedia.org/T375599


src/components/Footer/SmallPrintFooter.vue
```diff
@@ -0,0 +1,70 @@
...
+				<div class="wmde-banner-footer-item">
+					<span>Wikimedia Deutschland ist der gemeinnützige Verein hinter Wikipedia. Jede Spende ist steuerlich
+						absetzbar. Eine Spendenquittung erhalten Sie von uns.</span>
+				</div>
+			</div>
+		</div>
+
+		<div class="wmde-banner-footer-bank">
+			<label class="wmde-banner-footer-bank-item account">{{ $translate('donation-account') }}:
+				<SelectionInput :value="'Wikimedia e. V.'"/>
```
Abban, line 42: It's super ugly in the html but if you remove the line break between the labels and the `SelectionInputs` here it gets rid of the extra whitespace between them on the page. The compiler must be adding one or something.

It's not a blocker, but it might help save a teeny bit of horizontal space on smaller screens.

src/themes/Svingle/Footer/SmallPrintFooter.scss
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		&-usage {
+			color: #72777d;
+			padding: 5px 0;
+			margin: $left-column-margin;
+			text-align: left;
+			&-link,
+			&-link:visited {
+				color: #72777d;
```
Abban: This is a good chance to have a play with the swatches. You can add a `--footer-smallprint-color` variable into the light and dark colours and use it here with `var( --footer-smallprint-color )`.

You can optionally add a new flag to the mixin and the skin file to optionally include the variable, but that's up to you. I reckon we'll be using this footer for the campaign.

Abban (approved):
This is very good work. We'll have a chat with the PM about what to do on smaller sizes, and I've pinged him that the colour contrast is too low.

moiikana:
added a function that adds tracking to the FAQ link since last review

Sperling-0 (approved):

PR #556, opened 2024-09-30
C24_WMDE_Mobile_DE_05

moiikana:
- based on de 04 ctrl
- var has additional "thank you" text box

https://phabricator.wikimedia.org/T374103


src/components/ThankYouBox/ThankYouBox.scss
```diff
@@ -0,0 +1,14 @@
+@mixin layout() {
+	.thank-you-container {
+		height: 60px;
```
Abban, line 3: The height in the design is 40px

src/themes/Mikings/ThankYouBox/ThankYouBox.scss
```diff
@@ -0,0 +1,17 @@
+@use 'src/components/ThankYouBox/ThankYouBox';
+
+.wmde-banner {
+	@include ThankYouBox.layout;
+
+	.thank-you-container {
+		font-weight: bold;
+		background: #ebe2fc;
```
Abban: This is another thing we should create color variables for, and a flag I think.

But it also has me thinking, what do we do if they want to test 2 different colours of the box? We should chat about it in the daily tomorrow.

src/components/Icons/HeartIcon.vue
```diff
@@ -0,0 +1,9 @@
+<template>
+	<svg width="16" height="16" viewBox="0 0 16 16" fill="#6b4ba1" xmlns="http://www.w3.org/2000/svg">
```
Abban: When you put the colours into the swatch you can remove `fill="#6b4ba1"` and add `style="var( --thank-you-heart-fill )"`

Abban (changes requested):
It all looks pretty good, but I'm hitting request changes to remind us to talk about the colour variables in the daily tomorrow.

src/components/ThankYouBox/ThankYouBox.scss
moiikana: the design in the figma file is overall a bit different to our actual current live banners.
In figma, both, submit button and thank you box are 40px high.
In our live banners, the submit button has a height of 60px though.
That's why I assume I should keep them the same height (60px).

Abban: Ah right, grand so!

Abban (approved):

moiikana:
(rebased because of package-lock file)

PR #579, opened 2024-10-15
Add dark mode swatch to EN Mobile

Abban:
Enwiki Minerva now supports dark mode so we need to add a
dark swatch to the theme for the en mobile banner.

This also add swatching to the payment provider icons, which
is a backwards breaking change and may need to be integrated
into existing banners after being merged.

Ticket: https://phabricator.wikimedia.org/T376794


moiikana (changes requested):
Bildschirmfoto vom 2024-10-17 11-49-08
the highlighting text color is a bit wrong yet
the text color animation should probably be white (on dark) -> dark (on yellow)
but currently the text color animation is dark (on dark)->dark (on yellow)

moiikana (approved):

PR #583, opened 2024-10-16
Simplify layout of small print footer

Abban:
This removes flex styles and block level elements from
the small print footer which makes the line wrap a
little better on small sizes.

Ticket: https://phabricator.wikimedia.org/T375599


moiikana (changes requested):
Bildschirmfoto vom 2024-10-22 13-30-38

I'm not sure if something got changed or lost while rebasing but currently the footer misses some spacing

Abban:
Lets just close this, that footer isn't gonna be used again anyway.

PR #571, opened 2024-10-08
C24_WPDE_Desktop_01

gbirke:
Ticket: https://phabricator.wikimedia.org/T375609

PR #592, opened 2024-10-22
C24_WMDE_Mobile_DE_06

Abban:
Both banners now store the close choices in local storage
and send a new BannerSubmitOnReturnEvent if the local
storage item exists then a banner is submitted.

This is so we can finally check if the soft close does
anything.

VAR uses the Soft Close with an already donated button
that sets a 4 week timeout.

Ticket: https://phabricator.wikimedia.org/T375608


src/tracking/events/BannerSubmitOnReturnEvent.ts
```diff
@@ -0,0 +1,14 @@
+import { TrackingEvent, TrackingFeatureName } from '@src/tracking/TrackingEvent';
+
+export class BannerSubmitOnReturnEvent implements TrackingEvent<void> {
```
gbirke, line 8: I think this new event needs some explanatory comment on what it is for. E.g. "We fire this event when the donor submits the banner form after previously closing another banner. The 'userChoice' contains the name of the close choice (see CloseChoice enum) that the donor previously used."

src/components/DonationForm/MultiStepDonation.vue
```diff
@@ -33,11 +33,13 @@ interface Props {
 	stepControllers: StepController[];
 	pageScroller?: PageScroller;
 	formActionOverride?: string;
+	submitCallback?: () => void;
```
gbirke, line 37: For the sake of our future selves we should probably comment this. Somthing like "The component triggers this callback before it submits the form data to the donation page. You can use it for additional event tracking, but you should not use it for validation."

gbirke (commented):
Code looks good to me, just spotted two points where we might comment things because the names are not entirely obvious (but where I also could not come up with better names)

I'll copy the new `submitCallback` property of `MultiStepDonation` to #571 to track the fee.

src/components/DonationForm/MultiStepDonation.vue
Abban: You actually reminded me that I want to put it into the 2025 cleanup doc because I think it should be removed after the soft close test.

If the new fee thingy becomes standard (which I suspect it will) then we should check for it and fire that event in the donation form.

src/tracking/events/BannerSubmitOnReturnEvent.ts
Abban: When we finally have the new tracking schema we won't need this event at all, because we would put that information in as extra vars as part of the submit event

moiikana:
@Abban for the submit on return feature I suggest the following change to `LocalCloseTracker.ts`:
add the feature name and prepend it to the item in the local storage
```ts
import { CloseChoices } from '@src/domain/CloseChoices';
import hasLocalStorage from '@src/utils/hasLocalStorage';
import { TrackingFeatureName } from '@src/tracking/TrackingEvent';

export interface LocalCloseTracker {
	setItem( feature: TrackingFeatureName, closeChoice: CloseChoices ): void;
	getItem(): string;
}

export class LocalStorageCloseTracker implements LocalCloseTracker {
	public setItem( feature: TrackingFeatureName, closeChoice: CloseChoices ): void {
		if ( !hasLocalStorage() ) {
			return;
		}
		window.localStorage.setItem( 'fundraising.closeChoice', feature + '_' + closeChoice );
	}

	public getItem(): string {
		if ( !hasLocalStorage() ) {
			return '';
		}
		return window.localStorage.getItem( 'fundraising.closeChoice' ) || '';
	}
}

```

moiikana:
consider rebasing because https://github.com/wmde/fundraising-banners/pull/597 got merged

test/banners/mobile/components/BannerCtrl.spec.ts
```diff
@@ -128,6 +132,16 @@ describe( 'BannerCtrl.vue', () => {
 		} );
 	} );
 
+	describe( 'Soft Close Submit Tracking', () => {
```
gbirke: I was very confused by this name, because the banner has no soft close functionality. And you can see that only when you look at the test names. Maybe you could change the test name to `Soft Close Submit Tracking - No Soft Close`. And also remove the `skip` of the previous tests and delete it instead.

gbirke (commented):
Could you move the tests to `tests/banners/mobile/CS24_WMDE_Mobile_DE_06` to be prepared for #594 ?

test/banners/mobile/components/BannerCtrl.spec.ts
Abban: We made the decision to leave the tests in there with skips when we prepared the pre-campaign tests because it's easier to figure out what to put back in when we re-add the features.

gbirke (approved):

PR #599, opened 2024-10-29
Expose average donation to banner content

Abban:
We use the average donation to do the campaign projection.
This exposes it so it can be used in the content.

Also adds a banner test to make sure it works.

Ticket: https://phabricator.wikimedia.org/T378212


gbirke (approved):

gbirke:
I find it a little bit weird that the "average donation" is in the "Campaign projection" object where it could be confused with the "amount per minute" value. If you agree, you could add a note to the cleanup doc to move it out into the general CampaignParameters object.

Abban:
This is because that value was previously only used in the campaign projection. I'll put a into the cleanup and we can decide what to do with it next year.

PR #600, opened 2024-10-31
Fix - Mobile banner close button cut off on iphone

Sperling-0:
Ticket: https://phabricator.wikimedia.org/T378583


Sperling-0:
> [!NOTE]  
> Closing this PR because to test your custom branch on Shutterbug: "Name your branch in a way that the script can recognize which channel to use. The script does some string checks on the branch name: https://github.com/wmde/banner-screenshots/blob/main/src/CommandLine/determineCampaignFromBranchName.ts"

PR #601, opened 2024-10-31
Replace window timer with injected timer

Abban:
Components that use timers are hard to test. This
removed the window timers and replaces them with
an interface that we can mock.

Ticket: https://phabricator.wikimedia.org/T378589

### DIFF
--- a/banners/desktop/C24_WMDE_Desktop_DE_00/banner_ctrl.ts
+++ b/banners/desktop/C24_WMDE_Desktop_DE_00/banner_ctrl.ts
@@ -27,6 +27,7 @@ import { LocaleFactoryDe } from '@src/utils/LocaleFactory/LocaleFactoryDe';
 import { createFormItems } from './form_items';
 import { createFormActions } from '@src/createFormActions';
 import { currentCampaignTimePercentage } from './currentCampaignTimePercentage';
+import { WindowTimer } from '@src/utils/Timer';
 
 const date = new Date();
 const localeFactory = new LocaleFactoryDe();
@@ -72,5 +73,6 @@ app.provide( 'formItems', createFormItems( translator, currencyFormatter.euroAmo
 app.provide( 'formActions', createFormActions( page.getTracking(), impressionCount ) );
 app.provide( 'tracker', tracker );
 app.provide( 'currentCampaignTimePercentage', currentCampaignTimePercentage( date, page.getCampaignParameters() ) );
+app.provide( 'timer', new WindowTimer() );
 
 app.mount( page.getBannerContainer() );
--- a/banners/desktop/C24_WMDE_Desktop_DE_00/banner_ctrl.ts
+++ b/banners/desktop/C24_WMDE_Desktop_DE_00/banner_ctrl.ts
@@ -18,12 +18,8 @@ import { LocalImpressionCount } from '@src/utils/LocalImpressionCount';
 import { LegacyTrackerWPORG } from '@src/tracking/LegacyTrackerWPORG';
 import eventMappings from './event_map';
 import { createFallbackDonationURL } from '@src/createFallbackDonationURL';
-
-// Locale-specific imports
 import messages from './messages';
 import { LocaleFactoryDe } from '@src/utils/LocaleFactory/LocaleFactoryDe';
-
-// Channel specific form setup
 import { createFormItems } from './form_items';
 import { createFormActions } from '@src/createFormActions';
 import { currentCampaignTimePercentage } from './currentCampaignTimePercentage';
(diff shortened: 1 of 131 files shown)


test/components/BannerConductor/StateMachine/states/PendingState.spec.ts
```diff
@@ -19,26 +21,36 @@ describe( 'PendingState', function () {
...
+		let called = false;
+		let delay = 0;
+
+		const timerSpy = newTimerSpy();
+		timerSpy.setTimeout = ( fn: () => void, after: number ): number => {
+			called = true;
+			delay = after;
+			fn();
+			return 0;
+		};
```
gbirke: I'd prefer the `TimerSpy` to collect that information in the implementation (even if it's only needed in this test).

gbirke (approved):
This is a great change! Moving global functions with side effects to dependencies is looking very good.

banners/desktop/C24_WMDE_Desktop_DE_00/banner_ctrl.ts
```diff
@@ -27,6 +27,7 @@ import { LocaleFactoryDe } from '@src/utils/LocaleFactory/LocaleFactoryDe';
 import { createFormItems } from './form_items';
 import { createFormActions } from '@src/createFormActions';
 import { currentCampaignTimePercentage } from './currentCampaignTimePercentage';
+import { WindowTimer } from '@src/utils/Timer';
```
gbirke, line 26: Could you move these imports up in all banners (both `WindowTimer` and `currentCampaignTimePercentage`, to avoid misleading comments?

moiikana (changes requested):
WPDE desktop and mobile seem broken for me here,
as they use the BannerConductor they probably need to `provide` the Timer in their entrypoints too

Abban:
Good catch, thanks.

src/components/BannerConductor/BannerConductor.vue
```diff
@@ -77,6 +79,8 @@ async function closeHandler( closeEvent: TrackingEvent<void> ): Promise<any> {
 	await stateMachine.changeState( stateFactory.newClosedState( closeEvent ) );
 }
 
+onUnmounted( timer.clearAll );
```
gbirke: The Banner Conductor is never unmounted, currently the whole code stays in the DOM of the page, just hidden. All the timers continue to tick, leaking memory.

I can see three ways moving forward:
1. Make all components responsible to stop the timers they started (like `KeenSlider`). this can be confusing and hard, esp. if they are not unmounted but just hidden but the fact that they no longer displayed needs to reach them. Worst option IMHO.
2. Pass timer as a dependency to the "final" states of the state machine (closed & not shown) and call `timer.clearAll` there.
3. Add code to the "final" states of the state machines that allows to call [`app.unmount()`](https://vuejs.org/api/application.html#app-unmount) to thoroughly remove the banner and clear up its resources. Since [getting the app instance from a component is being removed from Vue](https://stackoverflow.com/q/75446100/130121), the best way to do this is probably via another injection, e.g. `app.provide( 'unmountBanner', () => app.unmount() )`

src/components/SoftClose/SoftClose.vue
```diff
@@ -58,33 +59,26 @@ const props = withDefaults( defineProps<Props>(), {
...
-			window.clearInterval( timer.value );
 			emit( 'timeOutClose' );
 		}
 	}, 1000 );
 } );
 
-onUnmounted( () => {
-	window.clearInterval( timer.value );
-} );
-
```
gbirke, line 89: I was baffled why you removed the `onUnmounted`, but then discovered the whole "BannerConductor is supposed to clear all timers but doesn't" issue.

gbirke (changes requested):
Wow, this update uncovered some flaws in our banner code, thanks again for doing this improvement! It'll benefit us and the donors in the future

gbirke:
I just discovered that there is one more change needed - injecting the timer in the WPDE banners

banners/desktop/C24_WMDE_Desktop_DE_00/banner_ctrl.ts
Abban: I think I'd rather remove the comments, the IDE auto imports are always gonna mess this up

src/components/BannerConductor/BannerConductor.vue
Abban: The timers should be cleared in the states anyway, as that's where the business logic for the conductor should be. I don't know why I put it into the components.

We should chat about cleaning up the banner properly on close though, as it would free up some browser memory

gbirke (approved):

moiikana (approved):

PR #609, opened 2024-11-06
Only lint CSS of active campaigns

gbirke:
Our StyleLint configuration was linting the growing `banners/`
directory. This commit helps to exclude all non-active banners, shaving
off about 5 seconds from the dev environment and linting task.


Abban (approved):

package.json
```diff
@@ -20,6 +20,7 @@
     "build:campaign": "node webpack/build_channel.js",
     "start": "npm run dev",
```
moiikana, line 21: if we support `start` and `dev`, we'd also need a `start-strict` :grimacing: :pray:

gbirke: Done

moiikana (approved):

PR #612, opened 2024-11-08
Add import sorting

Abban:
This adds an import sorting lint rule and autofixer.
Imports will be sorted alphabetically and the local
ones moved to the bottom of the list

Ticket: https://phabricator.wikimedia.org/T378589


gbirke (approved):
Sorry, I can neither enthusiastically approve or disapprove here.

I do like the fact that sorting the imports is taken out of the hands of us developers who might try to impose some kind of structure to them which then fails at the next IDE autoimport. The order does not help me particularly (I caught myself expecting an alphabetical order of the import objects, not of the import paths), but the more automation the better.

Except ... This adds 2.5 seconds to the linting process, (12 vs 14.5 on my machine). OTOH that's only for a full linting run when nothing is cached, so this mostly affects CI runs which do always have an empty cache.

So the best I can do here is an "I approve, but only if no one else rejects". So please wait for another review before merging.

Sperling-0 (commented):
I prefer the imports sorted by the component name or variable name being imported.
But I can train myself to have them sorted by paths. So, it's okay for me.

moiikana:
when I look at (or "briefly fly over") the imports list I usually look at the component names in the beginning first (to find something),
so for me only sorting by component name would lead to some improvement.

I'm not sure if that benefit would be able to compensate the +2,5sec linting duration for me

gbirke:
Closing this for now, there was not enough enthusiasm for input sorting

PR #621, opened 2024-11-14
Add script to check tracking configuration

moiikana:
- in the past it happens that we forgot to adjust all the test numbers/start dates for banners in the `campaigns_info.toml` file 
-->  This commit adds a node command that runs a script to check the configuration file for consistency on those 2 issues.

- cleans up some codestyle in related files
- creates a "tools" folder for add-on scripts
- updates documentation

>[!TIP]
you could include `npm run check-tracking-conf` in your local `git hooks` (e.g. pre-commit)


Sperling-0 (approved):
I tried it out. I like the feature. Having colorful response would be very nice (Green for Success, and Red for Faulty tracking config).

Sperling-0:
> I tried it out. I like the feature. Having colorful response would be very nice (Green for Success, and Red for Faulty tracking config).

> [!NOTE]  
>I added a commit for the colored response. I hope you don't mind.
Feel free to discard it if you don't like it.

moiikana:
> I tried it out. I like the feature. Having colorful response would be very nice (Green for Success, and Red for Faulty tracking config).

would be nice to give it a more "blocking" and hinting character, but I'm currently not sure if it's possible to properly achieve that with javascript scripts in bash

tools/check-tracking-number-and-date.mjs
```diff
@@ -0,0 +1,48 @@
...
+		!configValues.banners.ctrl.tracking.includes( testNumberSearchString ) ||
+		!configValues.banners.var.tracking.includes( testNumberSearchString ) ) {
+		console.log(
+			`\x1b[48;2;139;0;0m\x1b[37m Faulty tracking config (test number) for: \x1b[0m\x1b[48;2;139;0;0m\x1b[37m${configValues.campaign} \x1b[0m`
+		);
+		break;
+	}
+	else {
+		console.log(
+			`\x1b[48;2;34;139;34m\x1b[30mEverything is fine with tracking config of \x1b[0m\x1b[48;2;34;139;34m\x1b[30m${configValues.campaign}\x1b[0m`
```
gbirke: I'd like to follow the UNIX philosophy of "No output on success". Imagine multiple lines of success messages pushing away the error or warning messages of other checks. Multiple "green" lines can lure you into a false sense of security.

If you really want this, you could add a `--verbose` flag to the tool.

tools/check-tracking-number-and-date.mjs
```diff
@@ -0,0 +1,48 @@
...
+	}
+	testNumber = testNumberMatchResult[ 0 ];
+	const testNumberSearchString = testNumber.slice( 1 ) + '-';
+	if ( !configValues.tracking.includes( testNumberSearchString ) ||
+		!configValues.banners.ctrl.tracking.includes( testNumberSearchString ) ||
+		!configValues.banners.var.tracking.includes( testNumberSearchString ) ) {
+		console.log(
+			`\x1b[48;2;139;0;0m\x1b[37m Faulty tracking config (test number) for: \x1b[0m\x1b[48;2;139;0;0m\x1b[37m${configValues.campaign} \x1b[0m`
+		);
+		break;
```
gbirke: You could change a boolean value, e.g. `trackingErrorsFound` to `true` here and do `process.exit( 1 );` after the for loop to exit the program and fail CI if an error is encountered. Same for the start date check.

Or have a collection of formatted error messages in an array and do `if( errors.length > 0 ){ errors.forEach(console.log); process.exit(1); }`

tools/check-tracking-number-and-date.mjs
```diff
@@ -0,0 +1,48 @@
...
+		console.log( 'Cannot parse test number ( "_dd" format) for ' + configValues.campaign );
+		break;
+	}
+	testNumber = testNumberMatchResult[ 0 ];
+	const testNumberSearchString = testNumber.slice( 1 ) + '-';
+	if ( !configValues.tracking.includes( testNumberSearchString ) ||
+		!configValues.banners.ctrl.tracking.includes( testNumberSearchString ) ||
+		!configValues.banners.var.tracking.includes( testNumberSearchString ) ) {
+		console.log(
+			`\x1b[48;2;139;0;0m\x1b[37m Faulty tracking config (test number) for: \x1b[0m\x1b[48;2;139;0;0m\x1b[37m${configValues.campaign} \x1b[0m`
```
gbirke: Please use the `chalk` library for coloring output - it will adapt to the shell environment, omitting the color codes when the environment does not support it, e.g. when used inside a pipe. See the file `vitest.campaign.config.mjs` for examples

tools/check-tracking-number-and-date.mjs
```diff
@@ -0,0 +1,48 @@
+import { parse as parseTOML } from 'toml';
+import { readFileSync } from 'fs';
+import { campaignInfoToCampaignConfig } from '../webpack/convert_info_to_type.js';
+
+const campaignConfig = parseTOML( readFileSync( 'campaign_info.toml', 'utf-8' ) );
+const configObject = campaignInfoToCampaignConfig( campaignConfig );
+
+for ( const configEntry of Object.entries( configObject ) ) {
+	const configValues = configEntry[ 1 ];
```
gbirke:
```suggestion
for ( const configValues of Object.values( configObject ) ) {
```

If you're omitting the channel names anyway, you can make this simpler. Naming-wise I'd also suggest using `campaignConfig` instead of the rather generic `configValues` and renaming `campaignConfig` in line 5 to `rawConfigForAllCampaigns` or something similar.

gbirke (changes requested):
Thank you for this valuable tool and the restructuring!

As a general approach in the code of the script, I'd prefer if we did not do small-scale regex matching and then checking strings. Instead, we should create additional value objects in `campaign_config_types.ts` for campaign names and tracking values that parse the strings in their constructor (and throw errors if the pattern is "broken") to make the patterns and contents very explicit and semantic. This will also help make the comparisons more meaningful.

Example code for the script
```typescript
const campaignName = new CampaignName(configValues.campaign)
const ctrlTracking = new KeywordTrackingCode(configValues.ctrl.tracking)
const varTracking = new KeywordTrackingCode(configValues.var.tracking)
const campaignTracking = new CampaignTrackingCode(configValues.tracking)
if ( campaignName.testNumber !== varTracking.testNumber ||  campaignName.testNumber !== ctrlTracking.testNumber) {
  // error logging here
}
// Another example where the "business logic" for comparison is moved into the value objects (tell-don't-ask)
if ( !ctrlTracking.matchesDate( varTracking) || !ctrlTracking.matchesDate( campaignTracking ) ) {
 // error logging here
}
```

It would be even better to move the object creation into the `Campaign` class with getters:

```typescript
class Campaign {
    // Properties here ...

    get campaignTracking() {
        return new CampaignTrackingCode( this.tracking);
    }
}
```

But then you'd have to change or duplicate `campaignInfoToCampaignConfig` in TypeScript to create the proper classes, so that's entirely optional

tools/check-tracking-number-and-date.mjs
moiikana: I want to fail as early as possible because if the "parsing steps" already don't succeed, the "comparison steps" also should not be executed either.
So I would like to `exit(1)` within the loop even

tools/check-tracking-number-and-date.mjs
moiikana: I would also like to keep it free of success messages

moiikana:
> const campaignName = new CampaignName(configValues.campaign)
const ctrlTracking = new KeywordTrackingCode(configValues.ctrl.tracking)
const varTracking = new KeywordTrackingCode(configValues.var.tracking)
const campaignTracking = new CampaignTrackingCode(configValues.tracking)
if ( campaignName.testNumber !== varTracking.testNumber ||  campaignName.testNumber !== ctrlTracking.testNumber) {
  // error logging here
}
// Another example where the "business logic" for comparison is moved into the value objects (tell-don't-ask)
if ( !ctrlTracking.matchesDate( varTracking) || !ctrlTracking.matchesDate( campaignTracking ) ) {
 // error logging here
}


@gbirke 
I would have assumed that creating 4 objects ( each of the constructors has to parse in each loop the ""org-mob09-241118-ctrl" input string for example (to get the testNumber) + then doing the string comparisons slows this loop down terribly, but I can also do that.

tools/CampaignConfigurationCheck/CampaignName.mjs
```diff
@@ -0,0 +1,15 @@
+import chalk from 'chalk';
+
+export default class CampaignName {
+	/**
+	 * @param {string} campaignName
+	 */
+	constructor( campaignName ) {
+		const testNumberRegexResult = campaignName.match( /_(\d{2})/ );
+		if ( testNumberRegexResult === null ) {
+			console.warn( `${ chalk.red( 'CAMPAIGN CONFIGURATION ERROR:' ) } Cannot parse test number ( "_dd" format) for ${ campaignName } ` );
```
gbirke: When I described these objects I was thinking of "simple value objects", without logging side effects and dependencies on console libraries. Could you throw exceptions instead and catch them when constructing? It'll be even less code since you have the `process.exit` in one place.

tools/CampaignConfigurationCheck/CampaignName.mjs
```diff
@@ -0,0 +1,15 @@
+import chalk from 'chalk';
+
+export default class CampaignName {
+	/**
+	 * @param {string} campaignName
+	 */
+	constructor( campaignName ) {
+		const testNumberRegexResult = campaignName.match( /_(\d{2})/ );
```
gbirke:
```suggestion
		const testNumberRegexResult = campaignName.match( /_(\d{2})$/ );
```

You could also do a `split` and use the parts, to make this class more applicable for future uses and would allow for more thorough checking.

```
const campaignParts = campaignName.trim().split( '_' )`;
if ( campaignParts.length < 5 ) {
   throw new Error( `Campaign name '${campaignName}' does not follow pattern, must have at least 5 parts, separated with underscores` );
}
const campaignNumber = campaignParts.slice( -1 );
if ( !/^\d{2}$/.test(campaignNumber) ) {
    throw new Error( `Cannot parse test number "${ campaignNumber }" of campaign "${ campaignName }". Test number must be a zero-padded, 2-digit numeric string` );
}
// const campaignYear = campaignParts[0].substr(1);
// etc ...
```

tools/CampaignConfigurationCheck/KeywordTrackingCode.mjs
```diff
@@ -0,0 +1,30 @@
+import chalk from 'chalk';
+
+export default class KeywordTrackingCode {
+	/**
+	 * @param {string} trackingCode
+	 */
+	constructor( trackingCode ) {
+		const testNumberRegexResult = trackingCode.match( /^\D*(\d{2})([-_])/ );
```
gbirke, line 6: Here you could also split at the `-` sign and examine the length and different parts with `^` and `$` regex chars. I always get a bit uneasy when I see a Regex without these "string boundaries" because it theoretically it could match anywhere in the string. Also, using boundaries makes the regex much faster because the regex engine does not have to try out all possible matches

gbirke (approved):
Some last suggestions for making the "business logic" side more explicit & separating it from the output and program flow. But I'm already approving as "good enough".

Thank your for your comment about performance of classes vs simple regexes. It's always a tradeoff between performance and readability. maintainability and developer experience. In our case the performance, even with a loop, is not too terrible, since we'll only have about 8 iterations (=number of channels).
But I must admit I have no hard-and-fast rules for making the tradeoff

tools/CampaignConfigurationCheck/CampaignName.mjs
moiikana: split certainly has its usecase, but if I start checking that thoroughly I see myself checking each element that split returns (to make sure I really got the right one) and nobody added a "-2-" for a test re-run for example.
From my perspective this would have a very strict assumption on the format.

gbirke (approved):

PR #622, opened 2024-11-15
Rename variable

Sperling-0:


gbirke (changes requested):
I think the `wmde-banner` prefix is redundant and unnecessary. We use it for our CSS *class names* to avoid clashes with MediaWiki class names. But for the *variable names* inside the banner repository they are implicit.
The general idea of making the parameters more semantic is helpful, though.

PR #624, opened 2024-11-15
C24_WMDE_Mobile_DE_09

moiikana:
based on 08 VAR

var has 10 reasons popup
and tracks
- [x]   The overlay is displayed.
- [x]   One of the reasons was expanded (and which).

- [x]   The button at the bottom of the overlay is clicked.

https://phabricator.wikimedia.org/T379953


banners/mobile/C24_WMDE_Mobile_DE_09/MinimisedTracker.ts
```diff
@@ -0,0 +1,27 @@
+import hasLocalStorage from '@src/utils/hasLocalStorage';
+
+/**
+ * This feature is to track the minimised state and load the banner
+ * minimised on page change. It was removed from the ticket, but
+ * I'm leaving it here in case it gets re-added while I'm away.
```
gbirke: Please delete this file and all other files related to minimisation - those files were not cleaned up in test 08 because it might not have been clear which banner wins when test09 was created. But since the minimize feature does not seem to come back, let's remove unused files in this test.

banners/mobile/C24_WMDE_Mobile_DE_09/styles/MiniBanner_var.scss
```diff
@@ -0,0 +1,161 @@
...
+				&:hover,
+				&:focus {
+					text-decoration: underline;
+				}
+			}
+		}
+	}
+
+	&-slide-reasons-to-donate-sentence {
+		margin-bottom: 0 !important;
```
gbirke: `!important` should be the last, desparate tool you reach for after you have intensively tried out deeper and deeper selector specificity. In this case, I recommend moving the selector into the `.wmde-banner-mini` section above it, and omitting the `&` and using `.wmde-banner-slide-reasons-to-donate-sentence`

src/components/ReasonsToDonate/ReasonsToDonate.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<dialog class="wmde-banner-10-reasons" ref="reasonsDialogue">
 		<div class="wmde-banner-10-reasons-close">
-			<button @click="$emit( 'hide' )">
```
gbirke: I like the simpler event names that don't repeat the component name in them, they'll be easier to read in the template. But that's just my personal taste.

test/features/ReasonsToDonate.ts
```diff
@@ -0,0 +1,69 @@
...
+	expect( tracker.trackEvent ).toHaveBeenCalledOnce();
+	expect( tracker.trackEvent ).toHaveBeenCalledWith( new ReasonsToDonateShownEvent() );
+};
+
+const expectTracksReasonsToDonateCTAClickedEvent = async ( wrapper: VueWrapper<any>, tracker: Tracker ): Promise<any> => {
+
+	await wrapper.find( '.wmde-banner-reasons-to-donate-link' ).trigger( 'click' );
+	await nextTick();
+
+	console.log( wrapper.html() );
```
gbirke: Please remove this `console.log` when the test passes

test/features/ReasonsToDonate.ts
```diff
@@ -0,0 +1,69 @@
...
+
+	await wrapper.find( '.wmde-banner-10-reasons-cta' ).trigger( 'click' );
+	await nextTick();
+
+	expect( tracker.trackEvent ).toHaveBeenCalledTimes( 2 );
```
gbirke, line 35: The test should only expect 1.
This is probably due to the removal of the tracking on the "close" event inside the `ReasonsToDonate` component

gbirke (changes requested):

src/components/ReasonsToDonate/ReasonsToDonate.vue
Abban: I see that you're trying to make this more consistent with the Use of Funds, but I'd prefer if this component fired different events for the close and CTA click, as this removes the need for an enum, and the requirement for an if branch in all the banner components.

test/features/ReasonsToDonate.ts
```diff
@@ -0,0 +1,69 @@
...
+import { expect } from 'vitest';
+import { PageScroller } from '@src/utils/PageScroller/PageScroller';
+import { nextTick } from 'vue';
+import { Tracker } from '@src/tracking/Tracker';
+import { ReasonsToDonateShownEvent } from '@src/tracking/events/ReasonsToDonateShownEvent';
+import { ReasonsToDonateCTAClickedEvent } from '@src/tracking/events/ReasonsToDonateCTAClickedEvent';
+import { ReasonsToDonateItemClickedEvent } from '@src/tracking/events/ReasonsToDonateItemClickedEvent';
+import ReasonsToDonate from '@src/components/ReasonsToDonate/ReasonsToDonate.vue';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Abban: You can remove this and the second parameter from the function. When the later parameters aren't used, js doesn't care if the signature is different from the interface. Wild west typing.

test/features/ReasonsToDonate.ts
```diff
@@ -0,0 +1,69 @@
...
+};
+
+const expectTracksReasonsToDonateCTAClickedEvent = async ( wrapper: VueWrapper<any>, tracker: Tracker ): Promise<any> => {
+
+	await wrapper.find( '.wmde-banner-reasons-to-donate-link' ).trigger( 'click' );
+	await nextTick();
+
+	console.log( wrapper.html() );
+
+	await wrapper.find( '.wmde-banner-10-reasons-cta' ).trigger( 'click' );
```
Abban:
```suggestion
	await wrapper.find( '.wmde-banner-10-reasons-cta button' ).trigger( 'click' );
```

test/features/ReasonsToDonate.ts
Abban: The tracker (with the CTA button selector fix) fires 3 calls when the CTA is clicked:

- Open the dialog
- CTA click
- Mini banner expanded

We should probably not fire the last one, as we can assume the CTA event that we're tracking opens the main banner.

test/features/ReasonsToDonate.ts
```diff
@@ -0,0 +1,69 @@
...
+
+	expect( tracker.trackEvent ).toBeCalledWith( new ReasonsToDonateCTAClickedEvent() );
+};
+
+const expectTracksReasonsToDonateItemClickedEvent = async ( wrapper: VueWrapper<any>, tracker: Tracker ): Promise<any> => {
+
+	await wrapper.find( '.wmde-banner-reasons-to-donate-link' ).trigger( 'click' );
+	await nextTick();
+
+	await wrapper.find( '.wmde-banner-10-reasons-accordion-item' ).trigger( 'click' );
```
Abban: There's an example of what you need to do in the `ReasonsToDonate.spec.ts` tests
```suggestion
	wait wrapper.find( '.wmde-banner-10-reasons-accordion-item:nth-child(1) .wmde-banner-10-reasons-accordion-title' ).trigger( 'click' );
	await wrapper.find( '.wmde-banner-10-reasons-accordion-item:nth-child(8) .wmde-banner-10-reasons-accordion-title' ).trigger( 'click' );
```

Then check the nth-child number is passed to the event as it's close choice.

Abban (changes requested):
Some suggestions for fixing the tests

test/features/ReasonsToDonate.ts
```diff
@@ -0,0 +1,69 @@
...
+	await wrapper.find( '.wmde-banner-reasons-to-donate-link' ).trigger( 'click' );
+
+	expect( tracker.trackEvent ).toHaveBeenCalledOnce();
+	expect( tracker.trackEvent ).toHaveBeenCalledWith( new ReasonsToDonateShownEvent() );
+};
+
+const expectTracksReasonsToDonateCTAClickedEvent = async ( wrapper: VueWrapper<any>, tracker: Tracker ): Promise<any> => {
+
+	await wrapper.find( '.wmde-banner-reasons-to-donate-link' ).trigger( 'click' );
+	await nextTick();
```
Abban: You might not need to await the nextTick in these as the action caused by the button click should be performed in the previous one.

src/components/ReasonsToDonate/ReasonsToDonate.vue
moiikana: > I like the simpler event names that don't repeat the component name in them, they'll be easier to read in the template. But that's just my personal taste.

I hope I understood your point, but if we e.g. have 3 popup features that all send "hide" to the parent, 
I feared that the browser / DOM event handling could be messing up which event came from which component.
If that's not the case and events stay within their "contact way to the parent" then we can have simpler event names, yes.

test/features/ReasonsToDonate.ts
moiikana: nice, I thought the linter didn't like it yesterday

banners/mobile/C24_WMDE_Mobile_DE_09/content/BannerSlides_var.vue
```diff
@@ -0,0 +1,96 @@
...
+			Ihnen in diesem Jahr Wissen im Wert einer Tasse Kaffee geschenkt?
+		</p>
+	</KeenSliderSlide>
+
+	<KeenSliderSlide :is-current="currentSlide === 4">
+		<p>
+			Dann entscheiden Sie sich, eine der seltenen Ausnahmen zu sein, und geben Sie etwas zurück.
+			Falls Sie zögern,
+			<a
+				id="reasons-to-donate-link"
```
Abban: You can remove this id, it's appearing on the page twice

banners/mobile/C24_WMDE_Mobile_DE_09/content/BannerSlides_var.vue
```diff
@@ -0,0 +1,96 @@
...
+				class="wmde-banner-reasons-to-donate-link t-reasons-to-donate-link"
```
Abban:
```suggestion
				class="wmde-banner-reasons-to-donate-link"
```

Abban:
I created a follow up PR that adds explicit styles for the links in the content, and swaps the links to buttons https://github.com/wmde/fundraising-banners/pull/626

src/components/ReasonsToDonate/ReasonsToDonate.vue
Abban: Vue emits aren't generic browser events, and it binds the handlers directly (eg `@hide="doThisThing"`). So 3 components can all emit "hide" events, and as long as they're bound to different functions it'll be okay.

gbirke (approved):

Abban (approved):

PR #627, opened 2024-11-19
Move urgency message variable into config

Abban:
The campaign day sentence urgency threshold is currently
hardcoded.

This moves it into the campaign config.

Ticket: https://phabricator.wikimedia.org/T380269


src/page/PageWPDE.ts
```diff
@@ -73,6 +73,7 @@ class PageWPDE implements Page {
 	}
 
 	public getCampaignParameters(): CampaignParameters {
+		console.log( window.campaignParameters );
```
moiikana: needs to be removed

moiikana (changes requested):

moiikana (approved):

PR #636, opened 2024-11-22
C24_WMDE_Mobile_DE_10

moiikana:
- based on mobde09
- [x]  If the "Anderer Betrag" button is clicked in the mini banner, the amount option of €10 is changed to €5.

- var:
- [x]     When submitting the banner form with an amount below €10, users are redirected to a donation form variant.
- [x]     When submitting the banner form with an amount above €10, users are redirected to the default donation form.
- [x]     When submitting the banner form with any amount using direct-debit, users are redirected to the default donation form.

- [ ] tests added + fixed

https://phabricator.wikimedia.org/T380461


banners/mobile/C24_WMDE_Mobile_DE_10/banner_var.ts
```diff
@@ -42,7 +44,8 @@ const app = createVueApp( BannerConductor, {
 		useOfFundsContent: localeFactory.getUseOfFundsLoader().getContent(),
 		pageScroller: new WindowPageScroller(),
 		remainingImpressions: impressionCount.getRemainingImpressions( page.getMaxBannerImpressions( 'mobile' ) ),
-		localCloseTracker: new LocalStorageCloseTracker()
+		localCloseTracker: new LocalStorageCloseTracker(),
+		currencyFormatter: currencyFormatter
```
gbirke: Please don't pass the currencyFormetter as a prop, it's injected with `provide` (see line 63) and can be accessed with `const currencyFormatter = inject('currencyFormatter')` in any component. We're trying to use props only for reactive values, not for "services" that do something (return a result, have side effects).

banners/mobile/C24_WMDE_Mobile_DE_10/components/BannerVar.vue
```diff
@@ -194,6 +197,12 @@ const amountOptionsFive = localFormItemsBuilder.setAmounts( 5, 15, 25, 50, 100 )
 const amountOptionsTen = localFormItemsBuilder.setAmounts( 10, 15, 25, 50, 100 ).getItems().amounts;
 const amountOptionsForForm = ref<FormItem[]>( amountOptionsTen );
 
+const { formAction } = useFormAction(
+	inject<FormActions>( 'formActions' ),
+	10,
+	{ smallAmount: 'xf=0', largeAmount: 'ap=1' }
```
gbirke: should be `xf=1` to activate the "address-opt-out" feature. And we need to check with the PR to ask if we should just show the "normal" form for large amounts (which is `ap=0`( or if we should really show the special email-only form (`ap=1`)

gbirke (approved):

PR #637, opened 2024-11-22
[DNM] Animate text colour in highlighted text

Abban:
This is an example of how text colour can be
animated alongside our animated highlight


Abban:
Closing this sample

PR #638, opened 2024-11-22
[DNM] Add minimising banner sample

Abban:
This is a minimised banner for testing the
auto-minimise on scroll feature.

It can also be used as the basis for changes
to the upcoming minimised banner test.

Ticket: https://phabricator.wikimedia.org/T380464


Abban:
This is implemented

PR #641, opened 2024-11-26
Colorize error message to increase visibility

moiikana:
- content version check output looked like regular text in the console and can be overlooked


Sperling-0:
> * content version check output looked like regular text in the console and can be overlooked

Very useful. I also had the same thought ✨.

Sperling-0 (approved):

PR #642, opened 2024-11-27
C24_WMDE_Desktop_DE_17

moiikana:
- based on desktop 16
- var has minimise button and close button

https://phabricator.wikimedia.org/T380464


banners/desktop/C24_WMDE_Desktop_DE_17/components/BannerVar.vue
```diff
@@ -0,0 +1,219 @@
+<template>
+	<div class="wmde-banner-wrapper" :class="contentState">
+		<MainBanner
+			@form-interaction="$emit( 'bannerContentChanged' )"
+			v-if="contentState === ContentStates.Main"
+			:bannerState="bannerState"
+		>
+			<template #close-button>
+				<div class="wmde-banner-minimised-button-group">
```
gbirke, line 9: I find this class name prefix `wmde-banner-minimised` confusing, because those elements don't belong to the minimised banner. It's convenient to style in the `MinimisedBanner.scss`, but I'd rather extract this into its own component with its own state and give it better class names

gbirke (approved):

src/themes/Treedip/SoftClose/SoftClose.scss
```diff
@@ -12,7 +12,12 @@
 		&-columns {
 			padding: 8px;
 			flex-wrap: nowrap;
-			justify-content: center;
+			justify-content: space-between;
+
+			.wmde-banner-close {
+				position: static;
+				margin-right: -40px;
+			}
```
gbirke: This is now unnecessary with the change to `ButtonCloseWithText` which fixes the "components with same name" issue (by adding a double class selector).

gbirke (approved):
Since the minimize button will not come back, there is no need extract it into its own component or change the naming.

PR #643, opened 2024-11-27
C24_WMDE_Mobile_DE_11

Sperling-0:
Ticket: https://phabricator.wikimedia.org/T380854


gbirke:
Just a general comment: @Sperling-0 and me have been pairing on this and we're deliberately editing the `Mikings` theme to have a fast feedback cycle (and because it was not clear at the start that the changes for the new design involve more than just colors). We will probably turn the edited `Mikings` theme into a new theme and then reset `Mikings`.

Abban:
Yep, this is a case where it's a whole new theme. You could ask the PM to name it too.

src/themes/Mikings/DonationForm/MultiStepDonation.scss
```diff
@@ -19,15 +19,17 @@
...
 
 		&-button {
 			height: 60px;
 			line-height: 60px;
 			border-radius: 30px;
 			font-weight: bold;
-			color: var( --donation-form-button-color );
-			background: var( --donation-form-button-background );
+			font-size: 18px;
+			color: black;
```
Abban: It may seem like it's faster to add the colours inline here, but making them vars from the outset will save time and headaches in the long run. Going through a theme and replacing existing colours with vars isn't a fun job.

src/themes/Mikings/swatches/color_light.scss
```diff
@@ -9,6 +9,7 @@ $grey800: #202122;
 $grey700: #54595d;
 $grey400: #a2a9b1;
 $grey100: #eaecf0;
+$grey: #eeeeee;
```
Abban: These variables are named using the [semantic colour system](https://dev.to/ynab/a-semantic-color-system-the-theory-hk7) please stick to that. When first defined the vars were incremented by 100s because that gives room to insert new colours between existing ones so don't be afraid to add new ones between as a half step (eg $grey150).

Abban:
My suggestion for getting this passing and merged is to move the data attribute commit into a new PR, and change the close button class that was causing the issue during testing to match the other banners.

gbirke (approved):

PR #645, opened 2024-12-02
Add ADR for test markers

gbirke:
Suggestion for a separation of concerns between naming elements for styling and naming them for referencing them in tests.


gbirke:
As Abban suggested during the campaign, let's keep the ideas around this topic as an ADR (that we could revisit), but ultimately reject it and do nothing.

Please give a review as a signal that you had another look at the PR

moiikana (approved):

PR #650, opened 2024-12-06
[DNM] Add prototype thank you banner

Abban:
I had an idea for the thank you banner and
this is the prototype for it.


Abban:
This won't be used

PR #651, opened 2024-12-06
Fix body styles when modal is open

Abban:
The existing styles we were adding to the body
when we open a modal aren't working properly on
Vector 2022.

This sets the width of the body to ensure the
centre positioning does not get lost when the
body is set to fixed.


gbirke (approved):
I could not replicate the problem because there are no replication steps in the description, but I guess this is fine. Thanks for discovering this!

moiikana:
I also can't see any problem with the existing banners

Abban:
There's no banners that cause it yet

PR #652, opened 2024-12-06
C24_WMDE_Mobile_DE_12

Abban:
- Decrease form button text on small screens
- VAR uses a double progress bar

Ticket: https://phabricator.wikimedia.org/T381398


src/components/ProgressBar/ConfigurableDoubleProgressBar.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+	<div class="wmde-banner-double-progress" :style="{
+		'--wmde-banner-progress-bar-width': progressBarContent.percentageTowardsTarget + '%',
+		'--wmde-banner-progress-bar-time-width': currentCampaignTimePercentage + '%'
+	}">
+		<div class="wmde-banner-double-progress-labels" v-if="withExtraLabels">
```
gbirke, line 6: You could omit the `withExtraLabels` and [check for the existence of one of the label slots](https://vuejs.org/api/component-instance.html#slots) with `$slots['extra-label-left'] || $slots['extra-label-right']` instead.

gbirke (approved):

moiikana (approved):

PR #655, opened 2024-12-10
Improve coverage report

gbirke:
Bring up coverage by ignoring files that are not relevant for coverage.
This change helps us to spot files that *actually* lack coverage that
would otherwise be hard to find in a sea of "uncovered" files.

- Exclude config, factory and translation files from coverage
- Exclude development environment files from coverage.
- Exclude all banners not in current config (because older banners
  don't have individual tests). We should change this back in 2025 to
  always check coverage of all banners.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,8 @@
     "ci": "npm-run-all check-tracking-conf lint test check-content-version",
     "check-content-version": "ts-node tools/check-content-version.ts",
     "check-tracking-conf": "ts-node tools/CampaignConfigurationCheck/check-tracking-number-and-date.mjs",
-    "coverage": "vitest run --coverage",
+    "coverage:filtered": "vitest run --coverage --config vitest.campaign.config.mjs",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
     "lint:test-imports": "node test/find_mismatched_imports.mjs",
     "ci": "npm-run-all lint test check-content-version",
     "check-content-version": "ts-node check-content-version.ts",
-    "coverage": "vitest run --coverage",
+    "coverage": "vitest run --coverage --config vitest.campaign.config.mjs",
```
moiikana: maybe we can have something like 
```suggestion
    "coverage": "coverage:filtered",
    "coverage:filtered": "vitest run --coverage --config vitest.campaign.config.mjs",
    "coverage:all": "vitest run --coverage",
```
here as well, to not easily forget we do special filtering here

moiikana (approved):

PR #649, opened 2024-12-05
C24_WMDE_Desktop_DE_19

Sperling-0:
- `Both banners`:

  - The banner is based on the VAR banner of desktop-de-18
 
  - Emit the event `modalOpened` when the use of funds modal is opened.

  - Emit the event `modalClosed` when the use of funds modal is closed.
  
 - `Var`:

   - The variant banner design is changed according to the design file for the [slider](https://www.figma.com/design/GZ1lU5L8mpaLXvtwlNfMki/wikimedia-FUNDRAISING?node-id=3078-404&t=0TCV5fzWZENGGqLh-4) and the [non-slider](https://www.figma.com/design/GZ1lU5L8mpaLXvtwlNfMki/wikimedia-FUNDRAISING?node-id=3078-409&t=Cg3T7nNprwjxCnx8-4).
   
   - The variant banner integrates the "10 good reasons" feature using the sticker.
  
   - Diverging from the design, the additional element should be at the bottom (sharing the space with the slide indicators) to prevent overlapping.

   - Hovering over the new element
       -  turns the cursor into a pointer.
       -  adds underlining to the link text.

   - Interaction with `ReasonsToDonate` is tracked.

Ticket: https://phabricator.wikimedia.org/T380852

PR #657, opened 2024-12-11
Prepare C24_WMDE_Desktop_DE_19 for lateprogress switch

moiikana:
- [X] enable lateprogress in the local dev environment
- [X] desktop 19 var fix
 	- use ProgressBarAlternative to show donation goal to the right of the bar
- [x] desktop 19 fallback banner progress bar fix

fixup for https://github.com/wmde/fundraising-banners/pull/649

https://phabricator.wikimedia.org/T381957


gbirke (approved):

Sperling-0 (commented):
I user tested this locally, works successfully. Could we name the new progress bar something else than `ProgressBarAlternative`? Because ProgressBarAlternative can mean anything. I wouldn't remember how this component is different then the original progress bar!

moiikana:
> Could we name the ne wprogress bar as something else than ProgressBarAlternative?

Saw your comment too late, 
I'm sure we can in another PR, because I also don't know what "Alternative" actually means here

PR #660, opened 2024-12-12
Make name of test command more explicit

moiikana:
- the config file that is provided to the vitest command filters for the recent banner in each channel
- makes this more visible in the command name


gbirke (approved):
I can see your point, your change makes it more consistent.

But I don't like to type more and destroy my familiarity (I'm running "npm run test:banners" frequently from the command line). Would you be ok with keeping `test:banners` as an alias of `test:current-banners`?

PR #661, opened 2024-12-12
Fix Fallback Banner for EN Desktop

gbirke:
Test 06 introduced a border in the main banner which now needs to be overwritten in the Fallback banner since they share class names

### DIFF
--- a/banners/english/C24_WMDE_Desktop_EN_06/styles/FallbackBanner.scss
+++ b/banners/english/C24_WMDE_Desktop_EN_06/styles/FallbackBanner.scss
@@ -70,7 +70,7 @@ $breakpoint: 800px;
 			display: flex;
 			flex-direction: column;
 			background: var( --fallback-message-background );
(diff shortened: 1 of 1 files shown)


gbirke:
Already deployed to CN

moiikana:
Ticket: https://phabricator.wikimedia.org/T382001

moiikana (changes requested):
this change causes the ctrl banner to not have any border anymore
Bildschirmfoto vom 2024-12-13 15-46-26

gbirke:
Thank you for testing this so thoroughly and catching my silly mistakes!
I'll close this PR and won't fix the mistake (which would involve adding `Fallback_var.scss` to Desktop EN 06), because:
1. The banner will be replaced with (correct) the VAR tomorrow
2. Is unlikely to be revived.

I've created #675 to address the root cause of all the fallback banner problems in the cleanup.

PR #665, opened 2024-12-13
Add browser tab icon for banner dashboard

Sperling-0:
When many tabs are open, the banner dashboard tab gets easily lost, having a colorful emoji as a tab icon makes it easily recognisable


Sperling-0:
> [!NOTE]  
> I have chosen the greenish yellow ball as the dashboard tab icon because of its color. It's easy to spot it when many tabs are open. This (https://emojipedia.org/japanese-open-for-business-button) emoji was also in the competition, but I chose the ball because I really liked its color.
> Feel free to choose something else as tab icon if you don't like it.

moiikana:
I'm not against adding a tab icon / "favicon" for the dashboard.

I only know the convention of using `<link rel="icon" href="...." />` with an `.ico` file inside of the header though, like described in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link.
The `<link>` tag describes that there are (external) resources that belong to the document (web page), like e.g. stylesheets, icons, assets,...

I suspect that browsers are optimized for this convention in the way they render things like tabs, bookmarks (wherever the icon will be appearing).
If we want to keep it professional, I'd prefer to use the link tag, but I can understand that it's "just a dashboard" that we use for ourself and you wanted to keep the change small.

Abban:
You can use a png for a favicon these days too so there's no need to make a `.ico`. So you can grab the emoji you used and use that.

Sperling-0:
I have added the tab icon professionally now.
NVC Giraffe is our dashboard logo!
What is NVC Giraffe?

(This is my cue to inspire you to take the NVC training :)

Feel free to replace the favicon.

Abban (approved):

PR #666, opened 2024-12-16
Add visibility flag/toggle for drama text visibility

moiikana:
- can be used in the banner copy
- can be used in the progress bar labels

https://phabricator.wikimedia.org/T381955

### DIFF
--- a/banners/desktop/C24_WMDE_Desktop_DE_19/components/BannerCtrl.vue
+++ b/banners/desktop/C24_WMDE_Desktop_DE_19/components/BannerCtrl.vue
@@ -110,7 +110,7 @@ import { LocalCloseTracker } from '@src/utils/LocalCloseTracker';
 import { BannerSubmitOnReturnEvent } from '@src/tracking/events/BannerSubmitOnReturnEvent';
 import { Tracker } from '@src/tracking/Tracker';
 import { useBannerHider } from '@src/components/composables/useBannerHider';
-import BannerTitle from '@banners/desktop/C24_WMDE_Desktop_DE_15/content/BannerTitle.vue';
+import BannerTitle from '../content/BannerTitle.vue';
 
 enum ContentStates {
 	Main = 'wmde-banner-wrapper--main',
--- a/src/utils/DynamicContent/generators/ProgressBarContent.ts
+++ b/src/utils/DynamicContent/generators/ProgressBarContent.ts
@@ -10,6 +10,7 @@ export class ProgressBarContent implements DynamicProgressBarContent {
 	private readonly _donationSum: number;
 	private readonly _remainingDonationSum: number;
 	private readonly _isLateProgress: boolean;
+	private readonly _dramaText: string;
 
 	public constructor(
 		donationTarget: number,
@@ -18,7 +19,8 @@ export class ProgressBarContent implements DynamicProgressBarContent {
 		remainingDonationSum: number,
 		translator: Translator,
 		currencyFormatter: Currency,
-		isLateProgress: boolean
+		isLateProgress: boolean,
+		dramaText: string
 	) {
 		this._donationTarget = donationTarget;
 		this._percentageTowardsTarget = percentageTowardsTarget;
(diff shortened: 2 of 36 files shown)


src/utils/DynamicContent/generators/ProgressBarContent.ts
```diff
@@ -57,4 +60,8 @@ export class ProgressBarContent implements DynamicProgressBarContent {
 	public get isLateProgress(): boolean {
 		return this._isLateProgress;
 	}
+
+	public get dramaTextIsVisible(): boolean {
+		return this._dramaTextIsVisible;
+	}
```
gbirke:
```suggestion
	public get dramaText(): string {
		return this._dramaTextIsVisible ? this._translator.translate( 'drama-text' ) : '';
	}
```

banners/desktop/C24_WMDE_Desktop_DE_19/components/BannerCtrl.vue
```diff
@@ -110,7 +110,7 @@ import { LocalCloseTracker } from '@src/utils/LocalCloseTracker';
 import { BannerSubmitOnReturnEvent } from '@src/tracking/events/BannerSubmitOnReturnEvent';
 import { Tracker } from '@src/tracking/Tracker';
 import { useBannerHider } from '@src/components/composables/useBannerHider';
-import BannerTitle from '@banners/desktop/C24_WMDE_Desktop_DE_15/content/BannerTitle.vue';
+import BannerTitle from '../content/BannerTitle.vue';
```
gbirke, line 113: Great catch, we might add a custom linter code check in 2025 to disallow `@banner`-absolute imports in the banner directory itself (the import alias is mostly for testing convenience IIR) and only allow `@src`-absolute and relative imports inside the same banner directory.

gbirke (changes requested):
I think this change should be improved upon in three ways:

1. **Move text generation**. The `DynamicCampaignText` and `DynamicProgressBarContent` are the classes that should provide localized and translated texts, based on their input. Therefore they should both provide a translated "dramaText" property (that can be an empty string), based on their constructor parameters. This moves the logic from the components into the class that's responsible for content creation. It might also make our tests better - the test  for the logic becomes a quick unit test and we can have a more formalized "content supports dramaText" feature test where we can compare the slider/message contents with and without the drama flag and if they are the same, fail the test. Putting the translations in `src/utils/DynamicContent/messages/DynamicCampaignText.de.ts` will help with the separation of concerns and possible text changes/overrides in the campaign.
2. **Implement text generation in progress bar classes**:  Currently, you'd also have to duplicate the logic you put in `BannerTitle` into all progress bar components. As far as I can see the progress bars themselves don't do anything yet with the drama text (flag or text), so that's also something that should be implemented in this PR. 
3. **Separate concerns**: I'd put the text generation logic/property in _both_ classes `DynamicCampaignText` and `DynamicProgressBarContent`, so they can be used independently (and maybe even have different text translations). I don't like how the the current implementation in `BannerTitle` breaks the separation of concerns. It's not bad and somehow conclusive, since `progressBarContent` *is* a part of `dynamicCampaignContext`. Although we might question why and if that's just for a convenient setup in the entry points where we might be served better with a factory that creates both objects separately. But that's a completely different discussion.

src/utils/DynamicContent/generators/ProgressBarContent.ts
moiikana: > As far as I can see the progress bars themselves don't do anything yet with the drama text (flag or text), so that's also something that should be implemented in this PR.

as far as I understood this is done in banner tickets and not in this PR because there is no design definition or anything like that in the ticket

gbirke (approved):
Thanks for implementing my suggestions!
You are right about the progress bar implementations - with the new properties ready, we can postpone the implementation (and limit the amount of work) until they're actually needed.

I don't want to block this again, so I'm approving, but could you add 2 tests to `test/unit/utils/DynamicContent/DynamicCampaignText.spec.ts` (one for each flag state)?

PR #669, opened 2024-12-17
Add new Use of Funds 2024 modal

Abban:
- Add new use of funds components and styles.
- Add duplicate interface, environment imports and
  content loaders for the new content structure.
- Add tests.
- Add details to clean up document.

Ticket: https://phabricator.wikimedia.org/T381411


gbirke (approved):
I've user-tested it and found two bugs:
1. Content is missing. Please reset package-lock.json and use `npm run update-content` to update the new content.
2. The close button doesn't work. Console shows `Uncaught TypeError: payload is undefined` in the onCloseModal handler in the banner component

PR #670, opened 2024-12-17
C24_WMDE_Desktop_DE_22

Sperling-0:
`Both banners`:
- The banners are based on the control banner of desktop-de-21.

`Variant banner`:
- The variant banner shows an additional sentence.

Ticket: https://phabricator.wikimedia.org/T382301


banners/desktop/C24_WMDE_Desktop_DE_22/styles/styles.scss
```diff
@@ -12,7 +12,7 @@
 @use 'Banner';
 @use 'src/themes/UseOfFunds/UseOfFunds';
 @use 'MainBanner' with (
-	$banner-height: 357px,
+	$banner-height: 368.8px,
```
Abban: Giving this a fractional value will cause issues as OSs/browsers do subpixel rendering differently, and it's considered to not be a best practice in CSS. I would recommend leaving it at 357px, and reducing the font size of the text in the message in both banners.

Abban: You can actually look at the en banner for this, as that also has a load of text

Sperling-0: We had the slider break point increased at 1550 as suggested by the PM.

moiikana (approved):

PR #671, opened 2024-12-17
C24_WMDE_Desktop_EN_07

Abban:
Based on the VAR of test 06

VAR sends the donor to the ap=1 form page when amount
is above 10 or below and the donor wants a receipt.

Ticket: https://phabricator.wikimedia.org/T382304


src/components/DonationForm/SubComponents/SubmitValues.vue
```diff
@@ -4,6 +4,7 @@
 		<input type="hidden" name="amount" :value="formattedAmountForServer" />
 		<input type="hidden" name="interval" :value="interval" />
 		<input type="hidden" name="paymentType" :value="paymentMethod" />
+		<input type="hidden" name="receipt" :value="receipt" />
```
gbirke, line 7: I'm not particularly happy with this global change of `SubmitValues` (at least in its current, boolean form) because
1. It increases the size of the banner and the submitted data for all banners, even if none of them end up using that feature. Our PM has hinted that this might be the last test around this, so I'd have preferred a more localized instance of `SubmitValues`.
2. As far as I understand, the `receipt` parameter means "the user wants a receipt" to the application. Setting it to `false` by default means we would be changing/overriding an application setting from the banners (esp. how it's set up at the moment, see my comment https://github.com/wmde/fundraising-app-frontend/pull/548#discussion_r1890402180 ). If we decide to keep the global change to `SubmitValues`, the internal value should be nullable and null by default.

banners/english/C24_WMDE_Desktop_EN_07/useFormAction.ts
```diff
@@ -0,0 +1,21 @@
...
+import { useFormModel } from '@src/components/composables/useFormModel';
+
+export function useFormAction( formActions: FormActions, minimumAmount: number ): { formAction: Ref<string> } {
+	const formModel = useFormModel();
+	const formAction = computed( (): string => {
+
+		let URL: string = formActions.donateAnonymouslyAction;
+
+		if ( formModel.numericAmount.value >= minimumAmount || formModel.receipt.value ) {
+			URL = formActions.donateWithAddressAction + '&ap=1';
```
gbirke, line 12: A side observation, not a change request:
It always pains me to see string concatenation of URL parameters. That's what `URLSearchParams` is meant to do. Ill put in a wish for the  2025 cleanup doc for a more flexible `useFormAction` that allows for injection of a callback that receives FormValues and returns a URL. `useFormAction` should provide also some defaults/settings from the entry point (see `createFormActions`) and the default implementation of the callback contains would be a simplified version of `createFormActions` (that only switches out the URL part).

gbirke (changes requested):

src/components/DonationForm/SubComponents/SubmitValues.vue
Abban: You're right about defaulting it to it null, but IMO we now need the receipt option in the form:

- We've had multiple tests where we ask the donor about a receipt, and there's another desktop one coming https://phabricator.wikimedia.org/T382307
- I see the stakeholders wanting this option in banners again in the future as it helps them filter addresses while passing people who don't want to provide one straight to the payment provider.
- We've previously misused address type to judge if the donor wants a receipt, and this makes it more explicit.

banners/english/C24_WMDE_Desktop_EN_07/useFormAction.ts
Abban: IMO, we're limited atm by how we generate the URLs with their tracking as strings with `createFormActions`. We should replace those strings with 2 objects that we can append parameters to something like eg: `donateWithAddressAction.appendParameter( 'ap', '1' )` & `<form :action="donateWithAddressAction.url">`.

src/utils/FormModel/FormModel.ts
```diff
@@ -30,6 +30,7 @@ export interface FormModel {
 	disabledPaymentMethods: ComputedRef<string[]>;
 	addressType: Ref<string>;
 	addressTypeValidity: Ref<Validity>;
+	receipt: Ref<boolean>;
```
Abban: TODO: Fix this type
```suggestion
	receipt: Ref<boolean|null>;
```

gbirke (approved):

PR #672, opened 2024-12-18
Add single BC check for labeled PRs

moiikana:
- everytime a PR gets labeled with "Ready for review" the playbook job runs ( npm run ci and npm run test:all-banners)
- this could increase the response time to errors compared to scheduled checks per day (at midnight e.g)


gbirke (approved):
I still think that adding a label puts too much effort on us fallible and forgetful developers, but it is a good first step until we can hook up the actual review requests and run it only once

moiikana:
Bildschirmfoto vom 2024-12-18 18-31-17
I will ammend the commit and shorten this message

PR #684, opened 2025-01-21
Update device and plattform testing configuration

moiikana:
- we needed to update the test setup for the automated browser testing (testingbot), because in our configuration the browsers/devices were outdated or not supported by testingbot anymore

- this makes the setup more generic and easier to update

https://phabricator.wikimedia.org/T383402

### DIFF
--- a/campaign_info.toml
+++ b/campaign_info.toml
@@ -28,8 +28,8 @@ pagename = "B25_WMDE_Desktop_DE_00_var"
 tracking = "org-00-250114-var"
 
 [desktop.test_matrix]
-platform = ["edge", "firefox_win10", "chrome_win10", "safari", "firefox_macos", "chrome_macos", "firefox_linux", "chrome_linux"]
-resolution = ["800x600", "1025x768", "1280x960", "1600x1200", "1920x1200", "2560x1440"]
+platform = ["edge", "firefox_windows", "chrome_windows", "firefox_linux", "chrome_linux", "chrome_macos", "firefox_macos", "safari" ]
+resolution = ["800x600", "1280x720", "1536x864", "1920x1080", "2560x1440"]
 
 
 [mobile]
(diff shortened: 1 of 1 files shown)


campaign_info.toml
```diff
@@ -84,7 +84,7 @@ pagename = "B25_WMDE_iPad_DE_00_var"
 tracking = "org-pad00-250114-var"
 
 [pad.test_matrix]
-device = [ 'ipad_mini', 'ipad', 'ipad_pro_9_7_inch', 'ipad_pro_12_inch' ]
+device = [ 'ipad_mini', 'ipad_pro' ]
```
gbirke: There are 3 iPad sizes - ipad mini (8.3"), iPad (10,9") and iPad Pro/iPad Air (13"). 
See https://www.apple.com/de/ipad/compare/?modelList=ipad-pro-13-m4,ipad-10th-gen,ipad-mini-a17pro

Since one of purposes of the screenshots is discovering visual glitches at different resolutions, I'd be in favor of adding the "mid" size (named just `'ipad'`) here as well. I looked at the testingbot options - as long as we don't use iOS 18, we can use the iPad (10th Generation) as the mid-size model.

gbirke (approved):

campaign_info.toml
moiikana: yes, thank you. I'm still looking into mobile issues and will look into ipad next.

my current problem with the ipad setup is, that testingbot is giving mixed signals on what they provide.
on the "live testing" interface it seems very limited (no mini):
Bildschirmfoto vom 2025-01-21 13-12-42

in the documentation however they let you choose more ipads in "Specify Browsers&Devices":
https://testingbot.com/support/getting-started/php.html

moiikana: found answers to my questions on https://testingbot.com/support/getting-started/browsers.html and created device configurations for 3 different ipad sizes

moiikana:
dropped commit with ipad/iPad/pad unification attempt

gbirke (approved):

PR #686, opened 2025-01-24
Clean up donation form components

moiikana:
- Delete "upgrade to monthly interval" option, the feature is not used anymore by current banner tests
- Delete "UpgradeToYearlyForm" because it is unused
- Delete unused files and CustomAmountForm
- Delete unused StepControllers
- Remove TransactionFee feature 
- Delete ButtonCloseWithText feature 
- Document donation form features
- Clean up unused components in the donationform folder (+ test folders)

- [x] document features of remaining donation form steps and step controllers. maybe document which banner uses which form

https://phabricator.wikimedia.org/T383984


docs/Features/DonationForms.md
```diff
@@ -0,0 +1,82 @@
...
+
+- - - 
+## Main Donation Form Page
+
+### Basic Main Form
+- **file name(s):**
+  - `MainDonationForm.vue`
+- **description:**
+  - Basic Donation form that asks the user for interval, payment type and amount
+  - The button is set to `MainDonationFormButtonMultiStep.vue`, so it can dynamically change its label depending on invertal and payment type.
```
gbirke:
```suggestion
  - The button is set to `MainDonationFormButtonMultiStep.vue`, so it can dynamically change its label, depending on interval and payment type.
```

docs/Features/DonationForms.md
```diff
@@ -0,0 +1,82 @@
...
+  - `C25_WMDE_iPad_DE_00`
+  - `C25_WPDE_Mobile_00`
+  - `C25_WPDE_Desktop_00`
+
+### Dynamic amount options
+- **file name(s):**
+  - `MainDonationFormAdaptiveAmounts.vue`
+- **description:** 
+  - This form allows to pass in amount options for the donation form directly with a new property.
+  - It got introduced in `C24_WMDE_Mobile_DE_10`
```
gbirke: I'd like to be a bit more verbose here:

- The amounts no longer come from the `form_items.ts` file. Instead, the (mobile) banner component that includes `MainDonationFormAdaptiveAmounts.vue` has two sets of amounts and changes them, depending which button the user clicked in the mini banner. The idea behind this is ... (_let's ask PM if this is still active and important_)

If we decide to keep the feature, I'd like to create a ticket to move some of duplicated code from the banners into the form/formItems/etc

docs/Features/DonationForms.md
```diff
@@ -0,0 +1,82 @@
...
+- **used in (banner name):** 
+  - `C25_WMDE_Mobile_DE_00` 
+
+
+### ( Show Donation Receipt checkbox (generic) )
```
gbirke: We could mark the unused features with an emoji like ⬇️ or 👇 or 🗑️
They would also show up in an autogenerated table of contents

src/components/DonationForm/SubComponents/SubmitButtons/MainDonationFormButtonMultiStep.vue
```diff
@@ -5,6 +5,12 @@
 </template>
 
 <script setup lang="ts">
+/**
+ * This Button handels displaying different labels based on
```
gbirke:
```suggestion
 * This Button handles displaying different labels based on
```
a typo that repeats also in other components, plz search-and-replace

gbirke (approved):
I'd say let's merge this PR (after the small fixes I suggested) to have a stable basis to build on.

I like your documentation, also in the list format. I tried to imagine the lists as tables, but image that they would contain lots of whitespace and would be awkward to read on smaller sceens. 

One thing we should add to the documentation is one or two sentences per feature *why* it exists. This will help us in the future to decide to reevaluate assumptions and get into the mind of the campaigns team. Ideally, the *why* outlines the "research question" that the campaigns team had. It could also mention the most affected metrics (no. of donations, average donation amount, mean donation amount, no. of recurring donations, no. of non-anonymous donations, etc) to give us easy reference when it comes to "haggling" with the campaigns team when we wish to drop features.
But the "why" improvement can also be done in a followup PR.

For the different (and abandoned) receipt forms I'd say the current document is a great example on how we should document during the campaign. But for the coming campaign we need to consider our use case(s) for the documentation, the questions where we might look into the docs for answers. At the moment, I imagine the most frequent use case for us developers to be a short reminder "what is this feature and how does it work?", describing the current state. To answer this question I think it would be best to not do the "inheritance-style" docs, but have the current state in one place. That'll make the document shorter (imagine adding 5 new forms in the next campaign) and you don't have to jump around so much. We can add a "evolution of the feature" section that outlines the changes the the tests where the changes were made. That'll answer the questions that will also pop up, but less frequently: "When did this feature first come up?", "How much did we iterate on the feature"?
As a conseqence, I'd say we name the form and buttons plainly `MainDonationFormWithReceipt`. But we can do that in a separate PR.

PR #687, opened 2025-01-30
Clean up composables

moiikana:
- [x]  delete useAmountBasedDirectDebitDependentFormAction.ts and useAmountBasedFormAction.ts
- [x]  move useFormAction in DE Desktop to `src/components/composables` and name it `useFormActionWithReceipt.ts`
- [x] add tests for the `useFormActionWithReceipt.ts` composable
- [x] Use `useFormActionWithReceipt.ts` in EN banner, delete custom useFormAction in EN.
- [x] Document the feature "dynamic form action" in general and our two current variants (default form anction and the one from the ticket (See last tickets for EN/DE), someting about receipts and amounts. Document that the form action should be independent from the form and only rely on the form model.

https://phabricator.wikimedia.org/T383980


gbirke (approved):
Looking good as a cleanup.

I'd like to encapsulate URL generation a bit more, string-concatenating to an URL is not ideal. I'll create a new branch based off this.

PR #695, opened 2025-02-06
Create test banner for WPDE cookies

moiikana:
- we want to test whether setting the cookie with javascript instead of using image requests also works

https://phabricator.wikimedia.org/T384681


moiikana:
This banner setup was temporary for testing only

PR #708, opened 2025-02-12
Disallow absolute banner directory imports

moiikana:
- remove the `banners` alias from the config
- the absolute imports caused issues in banners because they were overlooked

https://phabricator.wikimedia.org/T384703


Abban (approved):
What were the issues being caused?

PR #714, opened 2025-02-18
Fix absolute path configuration

moiikana:
- add all paths that should be available in the test folder to tsconfig.test.json ("extending" the list of paths from the parent config file is not enough)

- fix for https://github.com/wmde/fundraising-banners/pull/708


moiikana:
tried to remove `@test` from the root tsconfig but that for some reason made the tests/imports which use `@test` fail.

test/tsconfig.json
```diff
@@ -2,7 +2,10 @@
 	"extends": "../tsconfig.json",
 	"compilerOptions": {
 		"paths": {
-			"@banners/*": [ "banners/*" ]
-		}
+			"@banners/*": [ "banners/*" ],
+			"@test/*": [ "test/*" ],
+			"@src/*": [ "src/*" ],
+			"@environment/*": [ "src/environment/prod/*", "src/environment/dev/*" ]
+		},
```
gbirke: Please remove the trailing comma here, otherwise strict JSON parsing might complain

gbirke (changes requested):

gbirke (approved):

PR #715, opened 2025-02-18
Introduce amountInCents

moiikana:
- use "amountInCents" instead of the float-y numericAmount

- use "minimumAmountInCents" as threshold value

- change submitvalues component to use the amountInCents value directly because
it's already getting the conversion to cents in `useFormModel`

- remove "numericAmount" as a field on FormModel

https://phabricator.wikimedia.org/T383982


src/utils/DynamicContent/formatters/CurrencyDe.ts
```diff
@@ -23,8 +23,18 @@ export class CurrencyDe implements Currency {
 		return this._euroAmountFormatter( amount ).replace( ',00', '' );
 	}
 
-	public customAmountInput( amount: number ): string {
-		return this._customAmountInputFormatter( amount );
+	public euroAmountFromCents( amountInCents: number ): string {
+		const floatValue = Math.trunc( amountInCents ) / 100;
```
gbirke, line 27: The `Math.trunc` is for removing any accidental decimal places that linger in the cent amount?

src/utils/parseIntegerFromFormattedString.ts
```diff
@@ -0,0 +1,20 @@
...
+	// Remove all non-numeric chars
+	amountStr = amountStr.replace( /[^0-9.]/g, '' );
+
+	const amountNumber = Number( amountStr );
+	if ( isNaN( amountNumber ) ) {
+		return 0;
+	}
+
+	// turn euro float into integer (cent) amount
+	return Math.trunc( amountNumber * 100 );
```
gbirke, line 19: Just FYI, I was about to suggest a different, mores string-based approach, also described in the ticket: splitting at the last dot, making sure the 2nd string always has 2 digits  concatenating the strings, and then convertig to number. In PHP that would have given us a proper integer.
However, TIL that [JavaScript stores *all* numbers as 64bit floats](https://medium.com/@vincentcorbee/lets-talk-numbers-in-javascript-72ffe1eb1f3f), so keeping this function almost identical to the one it replaces and just multiplying with 100 is much more sane and less error prone.

gbirke (approved):

src/utils/DynamicContent/formatters/CurrencyDe.ts
moiikana: exactly, the Math.trunc() is used for silently getting rid of wrong (float) input values.
Removing the method would call would cause an input of `1.99` (cents) to become `0,02 €` instead of `0,01€`

PR #722, opened 2025-02-24
Set maximum banner size to 250KB

Sperling-0:
Ticket: https://phabricator.wikimedia.org/T383997

### DIFF
--- a/webpack.production.js
+++ b/webpack.production.js
@@ -8,6 +8,8 @@ const LoadVueOnWpde = require( './webpack/load_vue_on_wpde' );
 const CampaignConfig = require( './webpack/campaign_config' );
 const path = require( 'path' );
 
+const BannerSizeWarningPlugin = require( './webpack/banner_size_warning_plugin' );
+
 function readWrapperTemplate( name ) {
 	// eslint-disable-next-line security/detect-non-literal-fs-filename
 	return fs.readFileSync( `./webpack/wikitext_templates/${name}.hbs`, 'utf8' );
@@ -83,7 +85,8 @@ module.exports = ( env ) => {
 							rimraf.sync( './dist/*.LICENSE.txt' );
 						} );
 					}
-				} )()
+				} )(),
+				new BannerSizeWarningPlugin( { limit: 250 * 1024 } ) // compiled banner size (in bytes)
 			],
 			externals: [
 				/**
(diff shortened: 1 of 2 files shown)


webpack/banner_size_warning_plugin.js
```diff
@@ -0,0 +1,60 @@
...
+							`\n\nDear FUN team, \nThe following banners exceed the banner size limit`,
+							`(${ this.banner_size / 1024 }KB) after compilation:\n` );
+
+						banners.forEach( banner =>
+							console.warn( `- ${ banner.name }:`,
+								chalk.red( `${ ( banner.size / 1024 ).toFixed( 2 ) }KB` ) )
+						);
+
+						console.warn( `\nFollowing features make it happen:\n---Features list---`,
+						`\n\nCould we please remove or optimise this feature?` );
```
moiikana: I like it, it's very precise :+1: 

Except the last sentence, where I think they probably can't answer this question to us very well (they don't know how to technically optimise, they rather would give up old features than the new current one, ...)
Maybe we could have a plea like 
"Please consider scraping older banner features or reducing the scope of the new feature."

webpack/banner_size_warning_plugin.js
```diff
@@ -0,0 +1,60 @@
+const chalk = require( 'chalk' ).default;
+
+class BannerSizeWarningPlugin {
+	constructor( max_banner_size ) {
+		this.banner_size = max_banner_size.limit; // limit in bytes
```
moiikana: (optional) 
we unfortunately don't have linting on those config/tooling files,
but usually use camelCase for variable names, so you could change it to camelCase in here as well

moiikana (changes requested):
(see my coment(s) above)

webpack.production.js
```diff
@@ -83,7 +85,8 @@ module.exports = ( env ) => {
 							rimraf.sync( './dist/*.LICENSE.txt' );
 						} );
 					}
-				} )()
+				} )(),
+				new BannerSizeWarningPlugin( { limit: 25 * 1024 } ) // compiled banner size (in bytes)
```
moiikana: probably caused by testing, remember to set it back to 250

Sperling-0: Fixed it. Sorry, will be careful of such things next time.

moiikana (approved):

PR #723, opened 2025-02-25
Separate fallback banner from main banners

Abban:
- Add fallback styles to component
- Add swatch and dark mode colours to fbb styles
- Create separate selectors for fbb components
- Create new banner for developing the fbb
- Hook up new styles and component in en and de banners
- Update tests
- Delete Fijitiv theme

Ticket: https://phabricator.wikimedia.org/T384698


moiikana (changes requested):
the english fallback banner has 2 red borders, should look like the fallback banner on the german channel

moiikana (approved):

PR #779, opened 2025-05-21
Add Event Tracking doc

Abban:
This adds a brief explanation of our banner tracking
module and a guide on how to set up tracking schemas
in the WMF infrastructure with links to the full
documentation and repositories.

Ticket: https://phabricator.wikimedia.org/T336842

### DIFF
--- a/docs/Event-Tracking-and-Schemas.md
+++ b/docs/Event-Tracking-and-Schemas.md
@@ -1,0 +1,101 @@
+# Event Tracking and Schemas
+
+In order to understand how well (or badly) a banner is performing we track donor events that are sent to external services. Common actions
+that are tracked include:
+
+- Banner was seen
+- Banner was not shown
+- Banner was submitted
+- A feature was interacted with
+
+If a new feature is added the performance of that feature is tested by adding it to a VAR banner and comparing it to a CTRL banner. We use
+tracking events to see if the feature is interacted with plus the overall performance of each banner is compared to see if there are changes
+in donor behaviour.
+
+## Domains
+
+We host banners in 2 different places, on wikipedia.org (ORG) and on wikipedia.de (DE). Banners hosted on DE use our self-hosted Matomo
+tracking, and banners hosted on ORG use the Wikimedia Event Platform.
+
+### Wikimedia Event Platform (WEP)
+
+This is a very powerful tracking tool that is used to track all events that happen on wikipedia.org.
+
+### Matomo
+
+This is an open source analytics tool. It is very basic compared to the Event Platform meaning we can't track as much as we do on the ORG
+banners, but it still works well enough for our purposes.  
+
+## Event Mapping
+
+Banners hosted on either domain fire the same internal events. These are bubbled up to the entry points, and they are then handled by the
+appropriate tracking code and sent to 1 of 3 external services. These services are:
+
+1. The 2025 ORG event schema on the WEP.
+2. The pre-2025 legacy ORG event schema on the WEP.
+3. The DE event schema on Matomo.
+
+### 2025 ORG Schema
+This is the schema our internal events were created for and these events map straight to the required structure.
+
+### Pre-2025 ORG Schema & DE Event Schema
+Both these schemas require different data structures from our internal events. In the folders of the banners that use these schemas you will find an `event_map.ts` file that handles mapping the internal events to external ones.
+
+### Schemas management on the Wikimedia Event Platform
+
+The Wikimedia Event Platform uses pre-defined schemas to track each event type. That means we need to define and deploy our own schemas
+in order to use it. This is done in 2 steps.
+
+#### 1. Add an [Event Stream Configuration Item](https://wikitech.wikimedia.org/wiki/Event_Platform/Stream_Configuration)
+
+Repository: https://gerrit.wikimedia.org/r/q/project:operations/mediawiki-config
+
+This lets the Event Platform know that the schema exists and where it's located. You will need to have [Gerrit](https://www.mediawiki.org/wiki/Gerrit/Tutorial)
+set up in order to make a PR there. We have an existing one and you most likely won't ever need to set a new one up.
+
+#### 2. Add (or modify existing) Schemas
+
+Repository: https://gitlab.wikimedia.org/repos/data-engineering/schemas-event-secondary
+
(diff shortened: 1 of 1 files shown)


docs/Event-Tracking-and-Schemas.md
```diff
@@ -0,0 +1,80 @@
...
+
+**Schema Cheat Sheet**
+
+Here is a small cheat sheet, but please refer to the documentation in the links above for full information on how to create and maintain these schemas.
+
+**Create a new Schema or Fragment**
+
+1. Add a new sub folder, in either in the fragments folder or base schema folder depending on what you're creating.
+2. Create a common.yaml file and edit it according to the [official documentation](https://wikitech.wikimedia.org/wiki/Event_Platform/Schemas#Creating_a_new_schema).
+3. When the common.yaml file is ready run `npm run build-new /path/to/mt/common.yaml` and it will generate the other schema files for you.
```
gbirke:
```suggestion
2. Create a `current.yaml` file and edit it according to the [official documentation](https://wikitech.wikimedia.org/wiki/Event_Platform/Schemas#Creating_a_new_schema).
3. When the `current.yaml` file is ready, run `npm run build-new /path/to/mt/current.yaml` and it will generate the other schema files for you.
```

I looked at your GitLab MR and docs and they talk about a file called `current.yaml`. "common" is a fragment inside `current.yaml` that probably has to be included.

docs/Event-Tracking-and-Schemas.md
```diff
@@ -0,0 +1,80 @@
...
+
+**Modify an existing Schema or Fragment**
+
+1. Modify the common.yaml file in the schema you need to change.
+2. Make sure to bump the version number of the schema
+3. When the common.yaml file is ready run `npm run build-new /path/to/mt/common.yaml` and it will generate the other schema files for you. These files will have the new version number meaning backwards compatibility will not be broken, and old banners can still use the old version.
```
gbirke:
```suggestion
3. When the `current.yaml` file is ready, run `npm run build-new /path/to/mt/current.yaml` and it will generate the other schema files for you. These files will have the new version number meaning backwards compatibility will not be broken, and old banners can still use the old version.
```

docs/Event-Tracking-and-Schemas.md
```diff
@@ -0,0 +1,80 @@
...
+set up in order to make a PR there. We have an existing one and you most likely won't ever need to set a new one up.
+
+#### 2. Add (or modify existing) Schemas
+
+Repository: https://gitlab.wikimedia.org/repos/data-engineering/schemas-event-secondary
+
+This contains the configuration for all the events that the Event Platform handles. You will use this to either [create new schemas](https://wikitech.wikimedia.org/wiki/Event_Platform/Schemas#Creating_a_new_schema) or to [modify existing schemas](https://wikitech.wikimedia.org/wiki/Event_Platform/Schemas#Modifying_schemas).
+
+Our schemas consist of 2 parts, the base schema and fragments. Fragments are small sub-schemas that are included by and make up base schemas.
+Think of them like includes, our base schema is made up of various fragments.
```
gbirke:
```suggestion
Our schemas consist of 2 parts, the **base schema** and **fragments**. The base schema has no event "field definitions" on its own, but consists of metadata and a list of fragments. Fragments are small sub-schemas that define the "fields" (data points) of the tracking event. 
Think of fragments like includes.
```

The order of introduction of concepts was a bit confusing to me until i read the actual schema YAML in the MR, so this is a suggestion to switch the order around

docs/Event-Tracking-and-Schemas.md
```diff
@@ -0,0 +1,80 @@
...
+
+Our schema files are located in:
+
+- Fragments: jsonschema/fragment/analytics/wmde_fundraising/
+- Base Schema: jsonschema/analytics/wmde_fundraising/banner_event/web[^1]/
+
+[^1]: web is used because we may eventually show banners in an app and will need a separate base schema for it
```
gbirke, line 70: Thanks for this footnote!

docs/Event-Tracking-and-Schemas.md
```diff
@@ -0,0 +1,80 @@
...
+banners, but it still works well enough for our purposes.  
+
+## Event Mapping
+
+This separation requires two separate implementations, one for each external service. 
+
+Banners hosted on either domain fire the same internal events. These are bubbled up to the entry points, and they are then handled by the
+appropriate tracking code and sent to the external service.
+
+In the banner folders you will find an `event_map.ts` file that handles mapping the internal events to external ones.
```
gbirke: I remember us talking about that the shiny new event tracking system will make the mapping obsolete, which is also why we have those "temporary" `event_map.ts` files and why we put the events we send to event platform into the `LegacyEventTracking` folder.

As far as I understand, it would be a huge hassle to recreate our domain events as fragments and would "block externally" any banner ticket that tries to introduce a new tracking metric. Is that correct?

If it is correct, I suggest we centralize the mapping a bit more to avoid duplication and drift in the `event_map.ts` files and dropping the "Legacy" moniker.

docs/Event-Tracking-and-Schemas.md
```diff
@@ -0,0 +1,80 @@
+# Event Tracking and Schemas
+
+In order to understand how well (or badly) a banner is performing we track donor events that are sent to external services. Common actions
+that are tracked include:
+
+- Banner was seen
+- Banner was not shown
+- Banner was submitted
+- A feature was interacted with
```
moiikana, line 9: thanks for the example section

moiikana (approved):

docs/Event-Tracking-and-Schemas.md
```diff
@@ -0,0 +1,86 @@
...
+
+- Fragments: jsonschema/fragment/analytics/wmde_fundraising/
+- Base Schema: jsonschema/analytics/wmde_fundraising/banner_event/web[^1]/
+
+[^1]: web is used because we may eventually show banners in an app and will need a separate base schema for it
+
+**Schema Cheat Sheet**
+
+Here is a small cheat sheet, but please refer to the documentation in the links above for full information on how to create and maintain these schemas.
+
```
gbirke, line 75: You could add some info on 
- when to create a new schema (rarely or never, only if we introduce a completely new class of events?)
- when to create a new fragment (tracking a series of new fields that belong together?)
- when to modify a schema (does it have to be modified when the fragments are modified or only when adding a fragment?)
- when to modify a fragment (when some small detail/type changes, adding a new field that makes sense together with the others?)

This will hopefully give our future selves a hint on what to do.

gbirke (approved):

PR #792, opened 2025-06-25
Remove SMS box

Abban:
We can no longer take sms donations so this removes the
SMS component

Ticket: https://phabricator.wikimedia.org/T397694


gbirke (approved):
Wohoo! 290 lines gone!

PR #793, opened 2025-06-25
WMDE_FR_2025_Mobile_DE_01

Abban:
- The banners are based on the variant banner of mobile-de-14 from last year.
- The first slide is removed.
- The slide display duration is increased by 2 seconds.
- The progress bar is removed from the modal banner.
- Clicking the close button of the mini banner doesn't show the soft close, but logs a close event and prevents banner display for 7 days.
- The link text of the use of funds links is changed to "Warum spenden?"
- The 3 column grid in the form is changes to a 2 column grid.
- The payment type Sofort is removed.
- The payment interval "vierteljährlich" is added.
- A dark mode color schema is added to the banners.
- VAR has slightly different copy.

Ticket: https://phabricator.wikimedia.org/T397694


gbirke (changes requested):
I think you forgot to hook up the new content in the last commit, VAR and CTRL still have the same content.

From a review perspective it would also be nice to have the VAR content already in the CTRL commit, so you can see the changes while reviewing and don't have to do the diff in the IDE

moiikana (approved):

PR #805, opened 2025-07-17
WMDE_FR_2025_Mobile_DE_02

moiikana:
- both banners based on VAR mob 01
- new VAR has 'already donated' feature at the top of the mini banner

Ticket: https://phabricator.wikimedia.org/T398010


banners/mobile/WMDE_FR_2025_Mobile_DE_02/styles/MiniBanner_var.scss
Sperling-0, line 1: I think we don't use underscore for `.vue`files. We use it only for  `.ts` and `.scss` files.
For `.vue` files we use `PascalCase`.

Sperling-0: Sorry, only now I realize that this is a draft PR. I should have commented only after you have opened it for review.

Abban (approved):

banners/mobile/WMDE_FR_2025_Mobile_DE_02/styles/MiniBanner_var.scss
moiikana: I've often been switching between underscore and pascalcase for this in the recent years, 
I usually prefer the underscore because it helps with the visual separation from the other files here.
I would leave it like this for now but will keep it in mind next time to keep more consistency,
if this is ok for you

PR #811, opened 2025-07-28
Update documentation for UoF

Sperling-0:
Ticket: https://phabricator.wikimedia.org/T400126


gbirke (changes requested):
I'm not sure about having a year-suffixed page on meta wiki. If we need new, incompatibe UoF data in the middle of the campaign and update the page, we'd break all past and existing banners (or would have to deploy all currently running banners at once and still break past banners).

We should discuss with the PM how we will deal with that and then change the documentation accordingly. I see two possible ways:
1) Reuse last year's UoF until it changes. This can be confusing and breaks if there is more than 1 change per year.
2) Add a suffix, e.g,  https://meta.wikimedia.org/wiki/MediaWiki:WMDE_Fundraising/UseOfFunds_2023_DE_2 Can also be confusing

PR #821, opened 2025-08-21
WMDE_FR_2025_Desktop_DE_04

Sperling-0:
- Both banners are based on the variant banner of desktop-de-03.

- Variant banner uses a different copy.


banners/desktop/WMDE_FR_2025_Desktop_DE_04/styles/Banner.scss
```diff
@@ -0,0 +1,22 @@
+@use 'src/themes/Svingle/variables/fonts';
```
Abban:
```suggestion
@use 'src/themes/Treedip/variables/fonts';
```

moiikana (approved):

PR #823, opened 2025-08-25
WMDE_FR_2025_Desktop_EN_01

Sperling-0:
- **Both banners:**
    - The banners are based on the **CTRL** banner of **desktop-en-07** of 2024
    - Both banners do not have any progress bars.
    - Both banners use the title "Please make it €x yearly" in the upsell
    - Both banners do not have a soft close banner (like DE desktop).

- **VAR Banner:**
   - The variant banner uses a different the copy
    
Ticket: https://phabricator.wikimedia.org/T402811


banners/english/WMDE_FR_2025_Desktop_EN_01/content/BannerSlides.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+	<KeenSliderSlide :is-current="currentSlide === 0">
+		<p class="headline">
+			<strong>
```
gbirke: We should try to avoid unsemantic (i.e. formatting-only) markup (like `<strong>` tags for a whole paragraph) in our code.

We have a `.headline` style in the Slider CSS definitions in our theme. Please remove the `<strong>` tag in the Slides.

For the non-slider text, please add the `.headline` class to the `p` tag and define a `.headline` style inside the `&-message` rule in `MainBanner.scss`

gbirke (approved):
Code looks good, user testing looks good. Just a small suggestion for best practices from me.

The ticket has a comment change requests ...

banners/english/WMDE_FR_2025_Desktop_EN_01/content/BannerSlides.vue
Abban: I wouldn't necessarily say its unsemantic here. It conveys the meaning that that part of text is emphasised more than the others. What we're doing here is basically `<p class="strong">` only we called it `.headline`. Using a strong tag also means we have one less style to maintain.

Screen reader users don't like emphasis being announced though, and `.headline` is already part of our themes so lets keep it.

banners/english/WMDE_FR_2025_Desktop_EN_01/content/BannerSlidesVar.vue
```diff
@@ -2,27 +2,28 @@
 	<KeenSliderSlide :is-current="currentSlide === 0">
 		<p class="headline">
 			<InfoIcon fill="#990a00"/>
-			{{ liveDateAndTime.currentDate }}, {{ liveDateAndTime.currentTime }}: &#8220;Wikipedia still can't be
-			sold.&#8221; - An important update for readers in Germany.
+			{{ liveDateAndTime.currentDate }}, {{ liveDateAndTime.currentTime }} - “Wikipedia still can't be sold.” -
```
moiikana, line 5: I'm not sure what the reason was to remove `&#8220;` here and use “ instead, but maybe not that important..

moiikana (approved):

PR #841, opened 2025-09-23
WMDE_FR_2025_Mobile_DE_05

Sperling-0:
- Both banners:
    - The banners are based on the `CTRL` of mobile-de-04.
    - Remove the redundant "Was Ihre Spende bewirkt" in the expanded banner

- Control banner

    - Everything stays the same, no changes here

- Variant banner

    - There is new highlighting.
    
Ticket: https://phabricator.wikimedia.org/T404978


banners/mobile/WMDE_FR_2025_Mobile_DE_05/styles/BannerVar.scss
```diff
@@ -0,0 +1,62 @@
...
+	&--not-shown {
+		.wmde-banner-wrapper {
+			display: none;
+		}
+	}
+
+	&-text-highlight {
+		padding: 0 5px;
+		color: var( --animated-highlight-color );
+		background-color: #990a00;
```
gbirke, line 60: **issue:** color values should come from themes

This is introducing a color value in the banner CSS, which we're trying to avoid. Please introduce a `$red600` in the Modo theme instead and create a new variable called `--animated-highlight-background`

gbirke (changes requested):

banners/mobile/WMDE_FR_2025_Mobile_DE_05/styles/BannerVar.scss
Sperling-0: @gbirke Are we keeping this _static_ text highlight feature?
Because it was an experimental feature I kept it in `BannerVar.scss`, I also discussed it in the daily.
If it is not a permanent feature then we should not update the theme, right?

gbirke: You're absolutely right, let's leave the color value here.
Do you have an idea on how we can make sure that if the variant wins we put the color value in the theme? Maybe a comment?

gbirke (approved):

banners/mobile/WMDE_FR_2025_Mobile_DE_05/styles/BannerVar.scss
Sperling-0: @gbirke I would create a color like `$red600` in the theme as you mentioned, then I would use that color in a variable called `--static-highlight-background`.
The catch is, one has to remember to make this change in the theme. I will put it in my TODO for now and will make the change in the upcoming banner if `VAR` banner wins wins.

PR #866, opened 2025-10-28
Bundle dependabot updates

Sperling-0:
- Bump @vitest/coverage-v8 from 3.2.4 to 4.0.4
- Bump vitest from 3.2.4 to 4.0.4
- `test.workspace` option was removed in Vitest 4, use `test.projects` instead
- https://github.com/wmde/fundraising-banners/pull/864/files
- https://github.com/wmde/fundraising-banners/pull/863/files


vitest.config.mjs
```diff
@@ -7,7 +7,7 @@ export default defineConfig( {
 
 	test: {
 		globals: false,
-		workspace: [
+		projects: [
 			// The `just-the-defaults` directory is a "workspace" that does not exist.
```
moiikana:
```suggestion
			// The `just-the-defaults` directory is a "project" that does not exist.
```

Sperling-0: Oh, thank you for spotting that. I will also look for comments (and also links for documentation) from now on. Thanks a lot.

moiikana (approved):

PR #859, opened 2025-10-23
WMDE_FR_2025_Mobile_DE_07

Abban:
Based on CTRL of test 6, VAR has Soft Close with the setup to measure the effectiveness.

PR #869, opened 2025-10-30
WMDE_FR_2025_Mobile_DE_08

Sperling-0:
- Both banners

    - The banners are based on the control banner of mobile-de-07.
    - The banners don't store the close action in the local storage.
    - When submitting the form the value of the close action previously stored in the local storage is logged as an event.

- Control banner

    - Submitting the banner form forwards users to the default form layout (ap=0).

- Variant banner

    - Submitting the banner form forwards users to the new form layout (ap=2).

- Notes

    - This ticket requires deploying the Fundraising Application.


gbirke (commented):
I guess the tests for this are blocked by #859 ?
When you add the tests, remember to switch out the tests that check for soft close tracking *storage* with tests that check for reading the previously stored soft close choice and tracking it as an event on submit. See also my comment on the Desktop 09 banner, the test routines for the feature are in `test/features/SoftCloseSubmitTracking.ts`.

Everything else looks good

moiikana (approved):

PR #870, opened 2025-11-01
thank-you-2025

Abban:
We have a new accessible thank you banner with a smashing new design.

Ticket: https://phabricator.wikimedia.org/T408117


banners/thank_you/components/BannerCtrl.vue
```diff
@@ -17,8 +17,8 @@
 			@membership-without-amount="membershipWithoutAmount"
 		>
 			<template #subscribe>
-				<a :href="subscribeURL" @click.prevent="subscribe">{{ thankYouContent[ 'more-info' ] }}</a>
-				<a :href="useOfFundsURL" @click.prevent="useOfFunds">{{ thankYouContent[ 'use-of-funds' ] }}</a>
+				<a :href="subscribeURL" @click.prevent="subscribe"><small>{{ thankYouContent[ 'more-info' ] }}</small></a>
+				<a :href="useOfFundsURL" @click.prevent="useOfFunds"><small>{{ thankYouContent[ 'use-of-funds' ] }}</small></a>
```
gbirke: I'm not sure if adding `<small>` is a good accessibility improvement.
The intro in https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/small makes it sound like it's a semantic element, but the note clearly states that it's just presentational like `<i>` or `<b>`. 
I've searched the web and could not find any evidence of this being read by screen readers. I would be interested in knowing why you added it.
My recommendation would be to remove the tags and optionally add a `role="contentinfo"` to the links (although that role does not fit 100%).

test/banners/thank_you/BannerCtrl.spec.ts
```diff
@@ -115,37 +85,31 @@ describe( 'BannerCtrl.vue', () => {
...
+		expect( wrapper.find( '.wmde-b-progress' ).attributes( 'style' ) ).toStrictEqual( '--wmde-b-progress-width: 80%;' );
 	} );
 
 	it( 'redirects when membership buttons are clicked', async () => {
 		const location = { href: '' };
 		Object.defineProperty( window, 'location', { writable: true, configurable: true, value: location } );
 		const wrapper = getWrapper();
 
-		await wrapper.find( '.wmde-banner-full-cta-with' ).trigger( 'click' );
+		const ctaFiveEuroAmount = wrapper.find( '.wmde-b-cta > div:first-child button' );
```
gbirke: I like this pattern of naming elements by using a variable (not quite a [Page Object](https://martinfowler.com/bliki/PageObject.html) but in the same spirit), we should use that more as a team, even if it gives us more of the "naming things" hard problem.

banners/thank_you/components/BannerModal.vue
```diff
@@ -4,6 +4,7 @@
 			<CloseButton
 				class="wmde-u-sticky"
 				:thank-you-content="thankYouContent"
+				data-with-background
```
gbirke, line 7: I still think putting in `data` attributes just to create minor variations that are not due to state (i.e. the data attributes dependent on state) is a mis-application of CUBE CSS.

banners/thank_you/components/BannerDisclosure.vue
```diff
@@ -0,0 +1,49 @@
+<template>
+	<details class="wmde-b-disclosure wmde-c-flow wmde-b-prose" :id="id" ref="dialogue" @toggle="onDialogueToggle">
```
gbirke, line 2: This is a really nice example of progressive enhancement/JS and Vue working together

gbirke (approved):
I've rebased this to `main` and fixed the commits the broke stuff.

PR #900, opened 2025-12-29
Avoid printing banners

gbirke:
This change hides banners via CSS when printing

### DIFF
--- a/src/components/BannerConductor/banner-transition.scss
+++ b/src/components/BannerConductor/banner-transition.scss
@@ -2,6 +2,10 @@ $transition-easing: cubic-bezier( 0.555, 0.155, 0.14, 0.945 ) !default;
 
 #wmde-banner-app {
 	top: calc( -1 * var( --wmde-banner-height ) );
+
+	@media print {
+		display: none !important;
+	}
 }
 
 .wmde-animate-banner {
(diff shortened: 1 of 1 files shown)


Sperling-0 (approved):

src/components/BannerConductor/banner-transition.scss
```diff
@@ -34,24 +38,31 @@ $transition-easing: cubic-bezier( 0.555, 0.155, 0.14, 0.945 ) !default;
...
+/**
+ * This class gets added to the <body> element when we show the banner.
+ * For themes with absolutely-positioned MediaWiki elements, we need to push down the by the banner height.
+ * We only do that on screens, so printing will be unaffected by banners
+ */
 .wmde-show-banner {
-	#wmde-banner-app {
-		top: 0;
-	}
+	@media only screen {
```
moiikana: I've read about the intention of "only" here,
because  I thought `@media screen` as opposed to `@media print` above would have been a sufficient statement here.

The sources I found on stackoverflow talk about using only to hide the following CSS from older/less common  browsers:
https://stackoverflow.com/questions/8549529/what-is-the-difference-between-screen-and-only-screen-in-media-queries/14168210#14168210
Also in the mentioned Adobe docs blog article:
> In spite of this behavior, it's still recommended to prefix media queries with `only` if you want to hide the styles from other, less common browsers.

As far as I understand this we actually don't want to use `only` here.

Was there a particular reason for putting it here?

We don't have a huge user base with extremely old browsers but nevertheless maybe it is better to remove an unintentional `only` here to avoid future confusion about syntax.

gbirke: Short answer: wrong assumptions, trusting what I saw in the browser more than what I read on the internet.

Long answer: I read the same articles when I created this and felt weird to put `only` there. But at the time, just putting `@media screen` in the CSS did not resolve the issue (a large gap at the top of the wikipedia Wikipedia page when you turn on the print view in the inspector after hiding the banner). When I added `only`, the gap disappeared. So I *assumed* that the internet was lying and that `@media screen` was not sufficient somehow. I've now tested with different browsers on different platforms and saw that the change works fine without `only`. 

Thank you for catching that, I'll remove it.

moiikana: > But at the time, just putting @media screen in the CSS did not resolve the issue (a large gap at the top of the wikipedia Wikipedia page when you turn on the print view in the inspector after hiding the banner). When I added only, the gap disappeared. 

I already suspected something like that to have happened :D

moiikana (approved):

PR #984, opened 2026-07-09
Fix page jump on mobile banner close

Abban:
We were firing the modal close event when the
mini banner was closed, and this caused the
page to jump to the top. This adds separate
close handlers for the mini and full page
banners.

Ticket: https://phabricator.wikimedia.org/T431694

### DIFF
--- a/banners/mobile/WMDE_FR_2026_Mobile_DE_01/components/BannerCtrl.vue
+++ b/banners/mobile/WMDE_FR_2026_Mobile_DE_01/components/BannerCtrl.vue
@@ -1,11 +1,11 @@
 <template>
 	<div class="wmde-banner-wrapper" :class="contentState">
 		<MiniBanner
-			@close="() => onClose( 'MiniBanner', CloseChoices.Close )"
+			@close="() => onMiniClose( 'MiniBanner', CloseChoices.Close )"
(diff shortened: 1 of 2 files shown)


banners/mobile/WMDE_FR_2026_Mobile_DE_01/components/BannerCtrl.vue
```diff
@@ -164,7 +164,11 @@ watch( contentState, async () => {
...
 
-function onClose( feature: TrackingFeatureName, userChoice: CloseChoices ): void {
+function onMiniClose( feature: TrackingFeatureName, userChoice: CloseChoices ): void {
+	emit( 'bannerClosed', new CloseEvent( feature, userChoice ) );
+}
+
+function onFullPageClose( feature: TrackingFeatureName, userChoice: CloseChoices ): void {
 	emit( 'bannerClosed', new CloseEvent( feature, userChoice ) );
 	emit( 'modalClosed' );
 }
```
moiikana, line 174: - this change needs to be applied to the VAR banner component as well
- the variant banner for some reason still contains the progress bar at 2 places, maybe you can remove that as well

moiikana (approved):

PR #989, opened 2026-07-15
Add option to submit to new tab

Abban:
We want to allow banners to open the donation form
in a new tab when submitted. This adds the following
to allow that:

- Add submit opens in new tab option to the donation form
- Add new submitted state to banner conductors that will
  make sure the banner is hidden when the donor returns
  to the tab

Ticket: https://phabricator.wikimedia.org/T432196

### DIFF
--- a/src/components/BannerConductor/BannerConductor.vue
+++ b/src/components/BannerConductor/BannerConductor.vue
@@ -13,6 +13,7 @@
 			:bannerState="bannerState.stateName"
 			:bannerHeight="bannerRef?.offsetHeight"
 			@banner-closed="closeHandler"
+			@banner-submitted="submitedHandler"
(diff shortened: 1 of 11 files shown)


src/components/BannerConductor/BannerConductor.vue
```diff
@@ -85,6 +86,10 @@ async function closeHandler( closeEvent: TrackingEvent<void> ): Promise<any> {
 	await stateMachine.changeState( stateFactory.newClosedState( closeEvent ) );
 }
 
+async function submitedHandler(): Promise<any> {
```
moiikana, line 89: `submittedHandler()`

moiikana (approved):

